PR #3, opened 2020-04-18
Matsim parsers

fredshone:
This implements matsim format (.xml and .xml.gz) read and write functions (in parse and write modules respectivelly) to and from the core Population class. Modified plans can now be used by MATSim transport simulations and vice versa.

Implementing MATSim format activity plans has required a few changes to core. One of these is a **breaking change for existing notebooks/hacking around**:

The breaking change is for any notebooks using indexing of populations or households:
eg:
```
household = population[0]
person = household[0]
```
All indexes are forced to string, so that the above would actually be:
```
household = population['0']
person = household['0']
```
I will add a sensible UserWarning to this effect. I had to do it to maintain a schema of sorts. Blame MATSim.

More generally there is support now for setting locations with coordinates (Point) or link ids (str) (thanks a bunch MATSim) and times are stores as datetime objects, but for convenience the date is allways the 1st january 1900 (🚂 ). 

Also includes a dodgy implementation of `write.write_travel_diaries()` that definately needs work...


pam/parse.py
```diff
@@ -1,146 +1,249 @@
...
+					activities.append(destination_activity)
+
+			household.add(person)
+
+		population.add(household)
+
+	return population
+
+
+def read_matsim(
```
KasiaKoz, line 110: Does https://github.com/arup-group/pam#dev need updating following the merge of this PR?

PR #5, opened 2020-04-21
implements EOD (end of day)

fredshone:
This aint pretty - I have implemented a method to crop plans that are longer than 24 hours. AKA "plan.crop()":

But firstly - I have implemented a variable throughout called `EOD` or "end of day". 
`EOD = datetime(year=1900,month=1,day=1,hour=23,minute=59,second=59)`

This feels a bit hacky but i think is correct. The day starts at the stroke of midnight and runs until 23 hours 59 minutes and 59 seconds later. I am not bothered about the lost second.

The year is still 1900 🤷 

```
def crop(self, limit=datetime(1900, 1, 2, 0, 0)):
    """
    crop a plan to some limit (typically a day). Plan components that start after this
    time are removed. Activities that end after this time are trimmed. If the last component is  
    a Leg, this leg is removed and the previous activity extended.
    """
    for idx, component in list(self.reversed()):
	if component.start_time >= limit:
		self.day.pop(idx)
    # deal with last component
    if isinstance(self.day[-1], Activity):  # trim activity end time
	self.day[-1].end_time = EOD
    else:
	self.day.pop(-1)  # remove last leg
	self.day[-1].end_time = EOD  # trim activity end time
```


pam/activity.py
```diff
@@ -380,6 +383,23 @@ def simplify_pt_trips(self):
...
 
+	def crop(self, limit=datetime(1900, 1, 2, 0, 0)):
+		"""
+		crop a plan to some limit (typically a day). Plan components that start after this
+		time are removed. Activities that end after this time are trimmed. If the last component
+		is a Leg, this leg is removed and the previous activity extended.
+		"""
+		for idx, component in list(self.reversed()):
+			if component.start_time >= limit:
+				self.day.pop(idx)
```
KasiaKoz, line 394: out of interest, why did you use `pop` and not `del`?

tests/fixtures.py
```diff
@@ -1,10 +1,13 @@
 import pytest
 from shapely.geometry import Point
+from datetime import datetime
 
 from pam.core import Population, Household, Person
 from pam.activity import Plan, Activity, Leg
 from pam.utils import minutes_to_datetime as mtdt
 
+EOD = datetime(1900, 1, 1, 23, 59, 59)
```
KasiaKoz, line 9: you can define it once somewhere where it makes most sense and then import it from that module

KasiaKoz: I think I saw it defined 5 times..

KasiaKoz: otherwise, I like that you made this a global variable

tests/test_3_parse.py
```diff
@@ -59,3 +60,13 @@ def test_agent_pid_4_complex_tour(test_trips, test_attributes):
     population = load_travel_diary(test_trips, test_attributes)
     acts = [a.act for a in population.households['1'].people['4'].activities]
     assert acts == ['home', 'work', 'work', 'work', 'work', 'home']
+
+
+def test_crop_last_act(person_crop_last_act):
+    person_crop_last_act.plan.crop()
+    assert person_crop_last_act.has_valid_plan
```
KasiaKoz, line 67: are there tests for `pam.activity.Plan.is_valid`? (which `pam.core.Person.has_valid_plan` uses) if not, I wonder if this is a valid test?

pam/activity.py
fredshone: i prefer pop syntax. del looks like python2 to me. Not sure about performance

tests/fixtures.py
fredshone: Yes - i will update

tests/test_3_parse.py
fredshone: ok i will add tests for is_valid() :(

pam/activity.py
KasiaKoz: `pop` returns the value of the item in the list you just deleted so it's just a bit funny to see it on it's own and not being used but i doubt it makes much of a difference performance wise

elladahan: Looks good to me! It bother me a bit that we lose the last second of the day, as you might get weird results when you try to sum stuff up and it doesn't add up to 24 hours. Why can't we use midnight on the next day as the EOD?  Also, why do you use limit and EOD when they should be the same value? Shouldn't the function just use limit?

fredshone: I acknowledge that its a bit awkward but i know it will work for both travel plans and matsim in the short term. MATSim doesn't deal well with time (or i don't deal well with how matsim deals with time). In the past we've had MATSim give times like `24:00:00` which is a pain so perhaps i've over reacted but that is now impossible with EOD as it is 😌 .

Longer term i plan to keep an eye on this and if a we can generalise better then yeah - we can. For example we could generalise to plans longer than 24 hours.

tests/test_3_parse.py
fredshone: this is done

tests/fixtures.py
fredshone: done

pam/activity.py
```diff
@@ -1,4 +1,7 @@
 from .utils import minutes_to_datetime as mtdt
+from datetime import datetime
+
+EOD = datetime(1900, 1, 1, 23, 59, 59)
```
mfitz: Why `EOD` over the more explicit `END_OF_DAY`?

pam/activity.py
fredshone: ok @elladahan i slept on it and i think will work with the more traditional 24 hours long day. EOD is now set to `datetime(1900, 1, 1, day=2, minute=0, second=0)`

pam/activity.py
fredshone: laziness. i will change

PR #8, opened 2020-04-24
Travel Diary sampling

Theodore-Chatziioannou:
Added an option of randomly sampling the travel population when loading travel diaries. 
For example, parse.load_travel_diary(trips, attributes, sample_perc = 0.01) will only load 1% of the travelers, with trip frequency used as the probability weight.

### DIFF
--- a/pam/parse.py
+++ b/pam/parse.py
@@ -12,13 +12,14 @@ from .utils import datetime_to_matsim_time as dttm
 from .utils import get_elems, write_xml
 
 
-def load_travel_diary(trips_df, attributes_df):
+def load_travel_diary(trips_df, attributes_df, sample_perc = None):
 
 	"""
 	Turn standard tabular data inputs (travel survey and attributes) into core population
 	format.
 	:param trips_df: DataFrame
 	:param attributes_df: DataFrame
+	:param sample_perc: Float. If different to None, it samples the travel population by the corresponding percentage.
 	:return: core.Population
 	"""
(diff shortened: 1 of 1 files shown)


pam/parse.py
```diff
@@ -251,3 +255,18 @@ def selected_plans(plans_path):
...
+	We merge the trips and attribute datasets to enable probability weights based on population demographics. 
+
+	:params DataFrame trips_df: Trips dataset
+	:params DataFrame attributes_df: Population attributes dataset. 
+	:params float sample_perc: Sampling percentage
+	:params string weight_col: The field to use for probability weighting
+
+	:return: Pandas DataFrame, a sampled version of the trips_df dataframe 
+	"""
+	sample_pids = trips_df.groupby('pid')[['freq']].sum().join(attributes_df,how='left').sample(frac=sample_perc,weights=weight_col).index
```
fredshone, line 271: Neat but I think we can drop the attribute_df join - i don't see how we'd use it without major rework?

Theodore-Chatziioannou: The attribute_df join is not used right now,  however -and depending the question we are trying to answer- we may wish in the future to keep a consistent demographic mix after sampling (in which case we would pass as weights a list of fields (ie weights=['freq','inc','age'...]. I don't mind it taking out if you think we won't need it..

fredshone: i agree, i am merging

PR #9, opened 2020-04-26
Parse challenge

fredshone:
Extremelly convoluted methods for infering activities from travel diaries. This combined the original rorys and theo mods.

Does marginally better at the challenge (from 9 to now 10 passing). However it does this without hardcoding default start activity in edge cases as 'work' so in practice is probably much better.

The tradeoff is compute (and readability). So I've made this more complex parse optional (default at the moment). This would be good to test...

The big additions are `infer_activity_idxs` and `infer_activities_from_leg_purpose`. The rest is just tweaks.


notebooks/pam-getting-started.ipynb
```diff
@@ -1141,9 +1141,9 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python (pam)",
+   "display_name": "Python 3",
```
KasiaKoz, line 1144: you changed your jupyter notebook kernel from your pam virtual environment to your base python installation ?

PR #10, opened 2020-04-27
Kasia add simple policies

KasiaKoz:
Adds functionality to two classes in the `modify` module to facilitate the simple policies defined in this [spreadsheet](https://docs.google.com/spreadsheets/d/1FQMa7dLe2cv1NEZnbu5cZo3v07tKXINwvOaLQYoEp-M/edit#gid=0). The notebook [PAM_Simple_Policies.ipynb](https://github.com/arup-group/pam/blob/kasia-add-simple-policies/notebooks/PAM_Simple_Policies.ipynb) goes through how these policies are implemented using the changes to the classes below.

1. `HouseholdQuarantined`
    - added `person_based` flag to handle person based probability value. Defaults to previous implementation with `probability` being a likelihood of a household having to be quarantined. Computes the probability for a household given probability for a person to be affected. See diagram [here](https://docs.google.com/presentation/d/1vOP4pu9Ot3z57GpniQ5On61BUa688oNn4F-WPUoQlsM/edit#slide=id.g840f175bf7_1_7) for more info.

2. `RemoveActivity` (most of the changes)
    - Extends the class to deal with different levels of policies given different levels of probability, the three levels are:
        - Household
        - Person
        - Activity
    how the probabilities work together can be seen in this [table](https://docs.google.com/presentation/d/1vOP4pu9Ot3z57GpniQ5On61BUa688oNn4F-WPUoQlsM/edit#slide=id.g840f175bf7_1_29), though the implementation in code traverses the probability tree.

Caveats
1. Passing more than one activity passed in the list increases the likelihood of the person/household being affected i.e. for 1.3 in [spreadsheet](https://docs.google.com/spreadsheets/d/1FQMa7dLe2cv1NEZnbu5cZo3v07tKXINwvOaLQYoEp-M/edit#gid=0) we want to remove `['education', 'escort']` but the probability should refer to the `'education'` activity only and `'escort'` should be affected as the result of `'education'` being affected. This is not the case right now and both activities contribute in the same way to the overall outcome.
It also means singular `'escort'` activities could be removed in the process.
2. This one is obvious but in this iteration the probability given pertains to any person (satisfying attribute conditions if defined) or any activity (listed in `activities` when defining `RemoveActivity`); not differentiation between types of people or weighing activities differently. (if separable i.e. not `['education', 'escort']` then you can just define several `RemoveActivity` policies with different `p`s and set different person attributes with different `p`s)

There are tests for this but I don't know how good they are, @mfitz I would really appreciate your critique.

`RemoveActivity` is a chonker, the policy types and probabilities could be abstracted into classes, I would appreciate your thoughts on the abstraction, in particular @fredshone, the abstraction master.


pam/modify.py
```diff
@@ -51,28 +63,164 @@ def apply_to(self, household):
 class RemoveActivity(Policy):
```
fredshone, line 63: Abstraction thoughts:

I would prefer a base class and several different child classes for use cases rather than mega classes with loads of if statements. I elaborate:

The base methods for all modifiers are __init__(), sample(), and apply_change().

We apply these modifier classes at the hh level and apply modifiers in sequence:
```
for hh in population:
    for modifier in modifiers:
        policy.apply_to(hh)
```
We must keep an eye on how these might interact badly.
 
We have 3 sampling levels; hh, person and activity. We can either sample exhaustively by looping, eg:
```
for person in hh:
    for activity in person:
        if sampler(activity)...
```
The advantage of this technique is that we can have unit specific sampling. But in a lot of cases, as you show for HouseholdQuarantine, we can 'one-shot' sample. One shot is neat and quick but assumes that i) all ps are equal and ii) that the policy is only going to be applied once then the modifier is done. 

So at this point i recommend a base class with 4 child classes with different samplers: hh_sampler, person_sampler, activity_sampler and person_simple_sampler. There might conceivably also be a activity_simple_sampler. In this case simple means not sampling with differentiation between people or activities. The samplers could default to simple if parameterised at such.

There is further differentiation between modifiers based on how they apply changes. Trivially we can sample at any level but also apply at any level. Giving at least 9 permutations. We can apply with a scope the same as the sampler, for example:
```
for person in hh:
    if sampler(person):
        person.stay_at_home()
```
this is equivalent to applying at a smaller scope, eg:
```
for person in hh:
    if sampler(person):
        for act in person:
            del act 
```
The awkward case is sampling at a scope smaller than the application. For example if an activity is sampled to be removed but we also want to check for a remove a parent escort trip. The HouseholdQuarantine is a simpler version of this.

In the case of HouseholdQuarantine, once sampled, the change is applied and the iteration moves on. The reason why this might not work in all cases, is if a modification alters the item being iterated over and/or we wish to keep iterating over it. This already happens and is dealt with:

```    
def remove_activities(self, person, p):
        seq = 0
        while seq < len(person.plan):
            act = person.plan[seq]
            if self.is_activity_for_removal(act) and p():
                previous_idx, subsequent_idx = person.remove_activity(seq)
                person.fill_plan(previous_idx, subsequent_idx, default='home')
            else:
                seq += 1
``` 
More broadly, I thing there could be some really complex applys in the future so no point trying to structure for them all. However I think we can usefully add a base class with empty sample and apply methods and gather often used methods (like the one above):

tests/test_5_activity_policies.py
```diff
@@ -320,3 +351,472 @@ def test_home_work_home_education_home_removal_of_education_act():
...
+
+    policy_remove_higher_education.apply_to(household)
+
+    for pid, person in household.people.items():
+        assert len(person.plan) == 5
+        assert isinstance(person.plan.day[2], Activity)
+        assert person.plan.day[2].act == 'education'
+
+
+def test_remove_second_education_activity(mocker, home_education_shop_education_home):
```
mfitz, line 461: `test_remove_activity_policy_only_removes_selected_activities`, or something like that?

tests/test_5_activity_policies.py
```diff
@@ -320,3 +351,472 @@ def test_home_work_home_education_home_removal_of_education_act():
...
+    mocker.patch.object(modify.RemoveActivity, 'is_selected', side_effect=[False, True])
+
+    person = home_education_shop_education_home
+    policy_remove_education = modify.RemoveActivity(['education'], probability=1)
+    policy_remove_education.remove_individual_activities(person)
+
+    assert len(person.plan) == 7
```
mfitz, line 468: Again, we're asserting about the end state of the list of plans here, based on implicit assumptions about the state before applying the policy. I'd prefer to be more explicit - assert on our assumptions before applying the policy, which would make it clearer what *change* we expect.

tests/test_5_activity_policies.py
```diff
@@ -320,3 +351,472 @@ def test_home_work_home_education_home_removal_of_education_act():
...
+
+def test_is_activity_for_removal_activity_does_not_match_RemoveActivity_activities():
+    activity = Activity(act = 'other_activity')
+    policy_remove_activity = modify.RemoveActivity(['some_activity'], probability=0.5)
+
+    assert not policy_remove_activity.is_activity_for_removal(activity)
+
+
+@pytest.fixture()
+def Steve_Smith_Household():
```
mfitz, line 511: This fixture doesn't yield a household, only a person, so the name should change, I think?

tests/test_5_activity_policies.py
```diff
@@ -320,3 +351,472 @@ def test_home_work_home_education_home_removal_of_education_act():
...
+        probability=0.5)
+    steve = household.people['1']
+    hilda = household.people['2']
+    timmy = household.people['3']
+    bobby = household.people['4']
+    policy.apply_to(household)
+
+    assert len(steve.plan) == 7
+    for i in range(0, 7, 2):
+        assert isinstance(hilda.plan.day[i], Activity)
```
mfitz, line 631: Hilda?

tests/test_5_activity_policies.py
```diff
@@ -320,3 +351,472 @@ def test_home_work_home_education_home_removal_of_education_act():
...
+        ['work'],
+        policy_type='activity',
+        probability=0.5)
+
+    policy.apply_to(Household(1))
+    modify.RemoveActivity.evaluate_activity_policy.assert_called_once()
+
+
+def test_evaluate_activity_policy_selects_steve_for_individual_activity_removal(mocker, Smith_Household):
+    mocker.patch.object(random, 'random', side_effect=[0] + [1 for i in range(11)])
```
mfitz, line 614: You know you can do `[1] * 10` to get a list with 10 1s?

tests/test_5_activity_policies.py
```diff
@@ -320,3 +351,472 @@ def test_home_work_home_education_home_removal_of_education_act():
...
+        assert isinstance(hilda.plan.day[i], Activity)
+    assert [hilda.plan.day[i].act for i in range(0, 11, 2)] == ['home', 'escort', 'shop', 'leisure', 'escort', 'home']
+
+    assert len(timmy.plan) == 11
+    for i in range(0, 11, 2):
+        assert isinstance(timmy.plan.day[i], Activity)
+    assert [timmy.plan.day[i].act for i in range(0, 11, 2)] == \
+           ['home', 'education', 'shop', 'education', 'leisure', 'home']
+
+    assert len(bobby.plan) == 5
```
mfitz, line 645: It's not obvious why any of these assertions should be true - they're kind of magic assertions based on a lot of implicit intimate knowledge of the fixture being used by the test. Is there a way to be more explicit?

tests/test_5_activity_policies.py
```diff
@@ -320,3 +351,472 @@ def test_home_work_home_education_home_removal_of_education_act():
...
+        probability_level='household',
+        probability=0.5,
+        attribute_conditions={'age': age_condition_under_10},
+        attribute_strict_conditions=True)
+    policy.apply_to(household)
+
+    modify.RemoveActivity.remove_household_activities.assert_called_once_with(household)
+
+
+def test_household_policy_with_person_based_probability(Smith_Household, mocker):
```
mfitz, line 685: I might think about testing a boundary condition in one of these tests, where the random number generated is exactly equal to the probability value.

tests/test_5_activity_policies.py
```diff
@@ -320,3 +351,472 @@ def test_home_work_home_education_home_removal_of_education_act():
...
+    policy.apply_to(household)
+
+    modify.RemoveActivity.remove_household_activities.assert_called_once_with(household)
+
+
+def test_household_policy_with_person_based_probability_with_a_satisfied_person_attribute(Smith_Household, mocker):
+    mocker.patch.object(modify.RemoveActivity, 'remove_household_activities')
+    mocker.patch.object(random, 'random', side_effect=[0])
+    household = Smith_Household
+    # i.e. Bobby is affected and affects activities on household level
```
mfitz, line 704: Here, maybe assert that exactly one member of the household is younger than 10?

tests/test_4_household_policies.py
```diff
@@ -52,6 +52,17 @@ def test_apply_full_hh_quarantine(population):
...
 
+def test_apply_person_based_full_quarantine(population):
+    policy = modify.HouseholdQuarantined(1, person_based=True)
+    modify.apply_policies(population, [policy])
+    assert len(population.households) == 20
+    for hid, household in population.households.items():
+        assert len(household.people) == 2
+        for pid, person in household.people.items():
+            assert len(person.plan) == 1
+            assert isinstance(person.plan.day[0], Activity)
```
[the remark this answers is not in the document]
fredshone: agree

pam/modify.py
```diff
@@ -51,28 +63,164 @@ def apply_to(self, household):
...
+    just one condition in attribute_conditions
     """
 
-    def __init__(self, activities: list, probability):
+    def __init__(self, activities: list, probability,
+                probability_level='activity', policy_type='activity',
+                 attribute_conditions=None, attribute_strict_conditions=True):
         super().__init__()
 
         self.activities = activities
```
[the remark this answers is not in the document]
KasiaKoz: yeah will change all that in some way :)

tests/test_4_household_policies.py
```diff
@@ -52,6 +52,17 @@ def test_apply_full_hh_quarantine(population):
             assert isinstance(person.plan.day[0], Activity)
 
 
+def test_apply_person_based_full_quarantine(population):
+    policy = modify.HouseholdQuarantined(1, person_based=True)
+    modify.apply_policies(population, [policy])
+    assert len(population.households) == 20
```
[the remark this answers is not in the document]
KasiaKoz: i guess to check that no households disappear? I agree that it doesn't match the unit test's purpose though, I'll make a change

pam/modify.py
fredshone: @KasiaKoz sorry for the long comment. TLDR: maybe split into smaller classes then look for shared methods

KasiaKoz: lol, it's ok, i like the content

tests/test_4_household_policies.py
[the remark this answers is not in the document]
KasiaKoz: cool cool, will change these tests

tests/test_5_activity_policies.py
```diff
@@ -112,6 +113,36 @@ def home_education_shop_home():
     return person
 
 
+@pytest.fixture
+def home_education_home_university_student():
```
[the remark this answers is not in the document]
KasiaKoz: did you like the smith household? :D

tests/test_5_activity_policies.py
```diff
@@ -320,3 +351,472 @@ def test_home_work_home_education_home_removal_of_education_act():
     assert [p.act for p in household.people['1'].activities] == ['home', 'work', 'home']
     assert household.people['1'].plan[0].start_time == mtdt(0)
     assert household.people['1'].plan[-1].end_time == END_OF_DAY
+
+
+def test_home_education_home_attribute_based_activity_removal_strict_satisfies_conditions(
+        home_education_home_university_student):
+    household = Household(1)
```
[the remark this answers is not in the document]
KasiaKoz: cool cool

tests/test_5_activity_policies.py
```diff
@@ -320,3 +351,472 @@ def test_home_work_home_education_home_removal_of_education_act():
...
+    def age_condition_over_17(attribute_value):
+        return attribute_value > 17
+
+    def job_condition_education(attribute_value):
+        return attribute_value == 'education'
+
+    policy_remove_higher_education = modify.RemoveActivity(
+        ['education'],
+        probability=1,
+        attribute_conditions={'age': age_condition_over_17, 'job': job_condition_education},
```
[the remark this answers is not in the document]
KasiaKoz: I like this as an addition to what Fred is suggesting, I think it's going to be to significant change altogether so will do this in the next PR

tests/test_5_activity_policies.py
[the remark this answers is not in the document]
mfitz: Love those guys... :-P

tests/test_5_activity_policies.py
```diff
@@ -320,3 +351,472 @@ def test_home_work_home_education_home_removal_of_education_act():
...
+        attribute_strict_conditions=True
+    )
+
+    policy_remove_higher_education.apply_to(household)
+
+    for pid, person in household.people.items():
+        assert len(person.plan) == 1
+        assert isinstance(person.plan.day[0], Activity)
```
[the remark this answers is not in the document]
KasiaKoz: yup, i wrapped it all in a method that checks the plans and activities for a person
```
def assert_correct_activities(person, ordered_activities_list):
    assert len(person.plan) % 2 == 1
    for i in range(0, len(person.plan), 2):
        assert isinstance(person.plan.day[i], Activity)
    assert [a.act for a in person.plan.activities] == ordered_activities_list
    assert person.plan[0].start_time == mtdt(0)
    assert person.plan[len(person.plan)-1].end_time == END_OF_DAY
```
and used it in all tests throughout

tests/test_5_activity_policies.py
```diff
@@ -320,3 +351,472 @@ def test_home_work_home_education_home_removal_of_education_act():
...
+
+
+def test_home_education_home_attribute_based_activity_removal_strict_doesnt_satisfy_conditions(
```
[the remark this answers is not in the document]
KasiaKoz: I swear coming up with test names is the hardest bit to do sometimes 🙈

tests/test_5_activity_policies.py
```diff
@@ -320,3 +351,472 @@ def test_home_work_home_education_home_removal_of_education_act():
...
+        ['education'],
+        probability=1,
+        attribute_conditions={'age': age_condition_over_17, 'job': job_condition_wasevrrr},
+        attribute_strict_conditions=True
+    )
+
+    policy_remove_higher_education.apply_to(household)
+
+    for pid, person in household.people.items():
+        assert len(person.plan) == 5
```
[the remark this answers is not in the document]
KasiaKoz: yup done

tests/test_5_activity_policies.py
KasiaKoz: i knew there was a trick, i just couldn't remember what it was, thanks :D

tests/test_5_activity_policies.py
KasiaKoz: shoulda been stevo, fixed indirectly through my personal activity assertion method

tests/test_5_activity_policies.py
KasiaKoz: added assertions prior to changes + the new activity assertion method should makes this much clearer, let me know if that works

tests/test_5_activity_policies.py
KasiaKoz: interesting, will add a test

tests/test_5_activity_policies.py
KasiaKoz: added some other assertions on people's attributes where they are relevant too

PR #12, opened 2020-04-29
Od matrices

IseulSong:
This is a function to extract OD matrices from Population and save it in a csv file. Please let me know how you think! :)


IseulSong:
Thanks for the review Fred. The three minor changes have been made as you requested and I've added a Trello card to add tests to the code later.

PR #13, opened 2020-04-29
vulnerability probabilities by demographic group

Theodore-Chatziioannou:
This touches two tasks from the backlog: "stats for vulnerable population of London/UK" and "joint distribution sampler". The calculations and narrative/methodology are part of the "vulnerability_probabilities.ipynb" notebook.

I am trying to estimate probabilities of a person being perceived as belonging to a vulnerable group (based on their age and gender), with the purpose of supporting mobility constrains in PAM. I am using a Bayesian approach, with mortality risk being the prior probability and covid deaths statistics the evidence, while a scaling factor normalizes vulnerability to a max of 100% likelihood for people over 70 (based on government advice).       

@fredshone I believe part of this was in the things you were looking at, so happy to discuss how to embed it in the sampling algorithms (if you are happy with the approach)


Theodore-Chatziioannou:
> Comment - maybe it makes sense to apply the adjustment by gender? Given that the risk is different for men and women, at the moment men have 100% risk at 65+. Would be interesting viewing and sharing without the adjustment - a sort of comparison of gov policy to reality.

It is possible to plot the without-adjustment risks, just change the last cell to plot_probabilities(df,'p_risk_group') (mortality risk if deaths were to reach 60k). Men so far seem to be more vulnerable to the virus, and I chose to keep this differential by scaling all categories together.  

The sampling approach sounds good!

PR #14, opened 2020-04-30
adds a function to export all population data to csv files 

elladahan:
This is a function to create csv files containing populations, households, people, activities and legs. 
These files can then be used as an input to a dashboard tool (in this case Power BI) to create statistics. Attached is a dashboard POC to show what kinds of things can be done with these csv files. 
[Dashboard POC.pptx](https://github.com/arup-group/pam/files/4558443/Dashboard.POC.pptx)

NOTE: The build is failing at the moment because I didn't include any tests with my code and therefore code coverage is slightly below the threshold. I'm working with mfitz to resolve this. Please do send your review comments in the meantime.

### DIFF
--- a/pam/write.py
+++ b/pam/write.py
@@ -147,11 +147,6 @@ def write_matsim_attributes(population, location, comment=None, household_key=No
 	# todo assuming v1?
 
 def create_population_export(list_of_populations, export_path):
-    
-    """"
-    This function creates csv export files of populations, households, people, legs and actvities. 
-    This export could be used to share data outside of Python or build an interactive dashboard.
-    """
     populations = []
     households = []
     people = []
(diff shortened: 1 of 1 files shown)


pam/write.py
```diff
@@ -145,3 +145,118 @@ def write_matsim_attributes(population, location, comment=None, household_key=No
...
+        """
+        Create population export
+        """
+        populations.append(
+            {
+                'Scenario ID': id(population),
+                'Scenario name': population.name   
+            })
+
+        pathfilename = export_path+str('/populations.csv')
```
fredshone, line 173: For building path names please use `os.path.join(export_path, 'populations.csv')` 
This will make paths that work on mac/windows and linux

pam/write.py
```diff
@@ -145,3 +145,118 @@ def write_matsim_attributes(population, location, comment=None, household_key=No
...
+    locations = []
+    
+    for population in list_of_populations:
+
+        """
+        Create population export
+        """
+        populations.append(
+            {
+                'Scenario ID': id(population),
```
fredshone, line 169: So Scenario ID is a sort of unique id? Could you use an index instead? ie 1,2,3...

fredshone:
```
for idx, population in enumerate(list_of_populations):
...
```

pam/write.py
```diff
@@ -145,3 +145,118 @@ def write_matsim_attributes(population, location, comment=None, household_key=No
...
+        
+        for hid, pid, person in population.people():
+            d_to_append = {
+                'Scenario_Household_ID': str(id(population)) + str("_") + str(hid),
+                'Scenario_Person_ID': str(id(population)) + str("_") + str(pid),
+                'Scenario ID': id(population),
+                'Household ID': hid,
+                'Person ID': pid
+            }            
+            people.append({**d_to_append, **person.attributes})
```
fredshone, line 205: noice

pam/write.py
```diff
@@ -145,3 +145,118 @@ def write_matsim_attributes(population, location, comment=None, household_key=No
...
+                    'Household ID': hid,
+                    'Person ID': pid,
+                    'Origin': leg.start_location.area,
+                    'Destination': leg.end_location.area,
+                    'Purpose': leg.act,  
+                    'Mode': leg.mode,
+                    'Sequence': leg.seq,
+                    'Start time': leg.start_time,  
+                    'End time': leg.end_time, 
+                    'Duration': leg.end_time - leg.start_time
```
fredshone, line 231: this is ok - but note that there is an existing method for getting the durations of legs and activities: `leg.duration`. But it does same thing.

pam/write.py
fredshone: `enumerate()` is a useful trick when iterating through an item (like a list in this case) - it returns the index (starting at 0) and the item for each iteration.

PR #16, opened 2020-04-30
Attribute sampling

fredshone:
Quick implementation for sampling vulnerable people:

`samplers.attributes.bin_integer_transformer()` for binning integer values (ie age). Intended to be used in a 'transformation loop', ie `for person in population: person.attributes['bin'] = ...`
This could easilly be replaced by preprocessing input data while in tabular format - but in this form it can also be used in a policy implementation.

`samplers.attributes.discrete_joint_distribution_sampler()` for random sampling from joint distribution. Can be used in a loop, applied to a table or in a policy as above.

I added the required dicts for these functions to the end of the `notebooks.vulnerability_probabilities.ipynb` notebook.

I am not precious about the implementation - just wanted something up and running fast for kasia. I quite like holding the distribution structure as a dictionary because we can extend it to N dims...


tests/test_9_samplers.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def test_discrete_joint_distribution_sampler_michael(michael, cat_joint_distribution):
+    mapping, dist = cat_joint_distribution
+    assert attributes.discrete_joint_distribution_sampler(michael, mapping, dist) == False
+
+
+def test_discrete_joint_distribution_sampler_kasia(kasia, cat_joint_distribution):
+    mapping, dist = cat_joint_distribution
+    assert attributes.discrete_joint_distribution_sampler(kasia, mapping, dist) == True
```
KasiaKoz, line 71: i think the names could be better, i couldnt tell what this was testing until I followed all of the crumbs up to the punchline of `discrete_joint_distribution_sampler`

fredshone: kk

PR #18, opened 2020-05-04
Od matrices

mfitz:
Fixed up the test and did some renaming of things to be more pythonic - note that this may require you to change any notebook code you have that calls this function.


pam/od_matrices.py
```diff
@@ -1,8 +1,7 @@
 import pandas as pd
-from pam.activity import Plan, Activity, Leg
-from pam.core import Population, Household, Person
 
-def extract_od(population, path='OD_matrix.csv'): #notice the default value
+
+def extract_od(population, path):
```
fredshone, line 4: ok - agree better to not use default paths :)

PR #20, opened 2020-05-05
London worker classification

fredshone:
# Notebook

Adds a prelim [notebook](https://github.com/arup-group/pam/blob/london-worker-classification/notebooks/london-workers-classification.ipynb) for person classification as 'able to wfh' and 'key worker' - based on NOMIS employment stats for London. The methodology is pretty janky:

1. using SOC2010 level 2 employment categories (this is currently a big dictionary in the notebook):
1.1 manually map to lopops occupation categories
1.2 manually classify as  'able to wfh' and 'key worker'
2. based on london population for each category (gender, occ and work status (ft/pt):
2.1 wrangle tables so that frequencies are exposed for  'able to wfh' and 'key worker'
2.2 convert to probability based on frequency
2.3 build distribution dicts

eg:
`{'occ1': {'ft': {'male': 0.5, 'female': 0.4}...`

Note that the distribution for each point is binary, ie probability of being key worker.

I am very happy to take feedback on the mappings or the methodology - feels wrong at the moment.

# Code

Also made some small changes to the attribute sampler so that if it fails to find a key it returns False. This behaviour is controlled by the value `careful` and set as default to not be careful, ie to return False. I like this behaviour because then the input distribution dictionaries do not need to be exhaustive.

For example in the case of work from home, a person with work status `in education` would be missing from the distribution dictionary, rather than failing it is useful to return False.

However, this does facilitate silent fails if a distribution dict is miss specified.

### DIFF
--- a/notebooks/london-workers-classification.ipynb
+++ b/notebooks/london-workers-classification.ipynb
@@ -1,0 +1,489 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "ExecuteTime": {
+     "end_time": "2020-05-05T12:46:55.725347Z",
+     "start_time": "2020-05-05T12:46:55.482694Z"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
+    "import os\n",
+    "from collections import defaultdict"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Intro\n",
+    "\n",
+    "This notebook shows the methodology for building a 'work from home' and a 'key worker' sampler for the population of London. The sampler uses a distribution based on person occupation, gender and work status (full-time or part-time). \n",
+    "\n",
+    "## Methodology\n",
+    "\n",
+    "Firstly we manually define three mappings based on SOC2010 occupation categories (level 2):\n",
+    "\n",
+    "- occ_mapping: mapping from SOC2010 cats to the lopops occupation cats\n",
+    "- key_worker_mapping: mapping from SOC2010 cats to key worker cat {0: not key worker, 1: key worker}\n",
+    "- home_worker_mapping: mapping from SOC2010 cats to home worker cat {0: not able to WFH, 1: able to WFH}\n",
+    "\n",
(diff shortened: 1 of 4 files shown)


Theodore-Chatziioannou:
If I read this correctly (and I may be wrong), I think you need to apply either step 1 or 2 - but not both. 
Given occupation, the probability of being a key worker is the same regardless of gender: P(key_worker|occ)=P(key_worker|occ,male)=P(key_worker|occ,female). 
The different demographic composition of each occupation category is already represented in the demand matrix, so even if the same probabilities are applied by gender/ employment status, the resulting matrix will still capture the distribution which (I think) you are aiming for with step 2. 

You can also do the reverse (with step 2): find the demographic characteristics of key workers and calculate probabilities to apply (based on gender/employment status), but leaving occupation field outside the distribution dictionaries, as it would have already been embedded in the results.

Let me know of any thoughts.

fredshone:
@Theodore-Chatziioannou 

So my defence is that i am using the additional SOC2010 distribution data for gender and work-status to better assign the probability of being a key-worker (for example) within an occupation group.

So i guess i disagree that: P(key_worker|occ)=P(key_worker|occ,male)=P(key_worker|occ,female)
becasue within each occupation group (from lopops) there are multiple soc2010 codes which have a distribution of key/not key across gender for example.

Non mathematically: given person A who has `occ1` i see that they are one of N possible SOC2010 codes, some of which are labelled `key`, some not. I then sample based on gender and employment status to more sensibly assign `key` from within this group of SOC2010 codes.

A concern is that the distribution of gender and employment status across the `occ` categories will (probably) be different from teh lopops and SOC2010 data.

ie; P_lopos(male|occ) != P_soc2010(male|occ)

Which i think is agreeing with your point that "The different demographic composition of each occupation category is already represented in the demand matrix," - but I am using my mapping to soc2010 codes to deterministically sample from that distribution.

Does that make sense?

Theodore-Chatziioannou:
@fredshone 
Thanks Fred, I see what you mean now and makes sense to me. The grouping in the build_dist() function is not making the key worker designation dependent on demographics, but rather their occupation within an occupation group. Which is better to what I had in mind, so I 'm happy with the approach :) .

PR #22, opened 2020-05-06
Add more unit tests

mfitz:
- Added some tests around @elladahan's code to export a population to CSV files
- Moved the OD matrix stuff that @IseulSong committed; didn't realise until now that there was a placeholder for that in the `write` module
- Unfortunate amount of formatting change where PyCharm replaced tabs - sorry about that
- Found some tests for the `write` module that weren't actually asserting anything; added minimal assertions that the files they expect to see are created, and TODOs for adding assertions about the expected contents of those files.
- Couple of minor changes (fixes) to code under test in order to make the tests work


pam/write.py
```diff
@@ -235,14 +234,11 @@ def create_population_export(list_of_populations, export_path):
...
                     'Location': activity.location.area,
-                    'Purpose': activity.act,  
+                    'Purpose': activity.act,
                     'Sequence': activity.seq,
-                    'Start time': activity.start_time,  
-                    'End time': activity.end_time,  
-                    'Duration': activity.duration
-            })
-
-        pathfilename = os.path.join(export_path, 'activities.csv')
```
elladahan, line 245: Just a small question, what was wrong with the naming 'pathfilename'?

tests/test_8_write.py
```diff
@@ -1,78 +1,253 @@
...
+
+
+def test_writes_households_csv_file_to_expected_location(population_heh, tmpdir):
+    write_population_csv([population_heh], tmpdir)
+
+    households_file = "{}/households.csv".format(tmpdir)
+    assert os.path.exists(households_file)
+    with open(households_file) as file:
+        households_file_lines = file.readlines()
+        assert len(households_file_lines) == population_heh.count(households=True) + 1
```
elladahan, line 139: Is the +1 for the header row?

tests/test_8_write.py
```diff
@@ -1,78 +1,253 @@
...
+
+    people_file = "{}/people.csv".format(tmpdir)
+    assert os.path.exists(people_file)
+    index = 0
+    people_in_pop = list(population_heh.people())
+    with open(people_file) as csv_file:
+        reader = csv.DictReader(csv_file)
+        for row in reader:
+            expected_hid, expected_pid, expected_person = people_in_pop[index]
+            assert row['Scenario ID'] == '{}'.format(index)
```
elladahan, line 159: Don't understand what this is checking...

pam/write.py
mfitz: > Just a small question, what was wrong with the naming 'pathfilename'?

I wouldn't say "wrong", but PyCharm's linting highlighted it as a bad name; the more pythonic version would separate words with underscores.

tests/test_8_write.py
mfitz: > Is the +1 for the header row?

Exactly :-)

tests/test_8_write.py
mfitz: There is actually an error here, and this only works because there is only one population. Rather than using `index`, it should just be using the fixed value `0`. It's checking that the `Scenario ID` is a string representation of what is, essentially, the population index.

mfitz: I pushed a commit to fix the population ID at 0 (because we know we're only exporting a single population) @elladahan

elladahan: Just to check.... this test only exports one population, but actually the script allows you to pass in multiple populations and create csvs for all of the populations you pass it. Shouldn't the test check that instead of only one population?

mfitz: Yes, it would be good to test that too (in a separate test(s), I think). Will add that after merging this one though, if that's okay with you?

elladahan: Cool!

PR #19, opened 2020-05-05
Add complex policies (part 1)

KasiaKoz:
Loads of new classes which aim to do everything else they did before this PR plus being able to give more than one probability to a policy and those probabilities can be customised by using custom samplers like in https://github.com/arup-group/pam/pull/16 but returning the p value instead. This enables easier implementation of other modifier policy classes

## Attribute Filter:
- `PersonAttributeFilter` : Handles people/households satisfying conditions, which are defined in the same format as when they were being passed to `RemoveActivity`, instead of `strict_conditions: True/False` we have `how: any/all`

## Policy Level:
- `HouseholdPolicy` : affects all activities in people in household
- `PersonPolicy`: affects all activities in person
- `ActivityPolicy` : affects single activity

## Sampling Level:
- `SamplingProbability` : base class
- `SimpleProbability` : default for single number probabilities given to the Policy level - is applied on the same level as the policy level
   
Below three take either a number or a custom sampler function that returns the `p` value based on input `x, **kwargs`, where `x` is the input at the level of `xProbability` (e.g. `x=household` for `HouseholdProbability`) and `kwargs` are inputs to the custom sampler function (there are examples in [Complex Policy notebook](https://github.com/arup-group/pam/blob/kasia-add-complex-policies/notebooks/PAM_Complex_Policies.ipynb))

- `HouseholdProbability` : computes probability for a household given p number for household
- `PersonProbability` : computes probability for a household or person given p number for person
- `ActivityProbability` : computes probability for a household, person or activity given p number for activity

## Modifier Classes:
- `RemoveActivity` : Now only contains methods to remove activities, sampling/filtering personal attributes is handled by Sampling/AttributeFilter classes respectively
Placeholders for:
- `MoveActivity`
- `AddActivity`

## Usage:
`PolicyLevel(ModifierClass, number or SamplingClass(or list of them), PersonAttributeFilter(optional))`
More examples in [Complex Policy notebook](https://github.com/arup-group/pam/blob/kasia-add-complex-policies/notebooks/PAM_Complex_Policies.ipynb)

Diagram
classes_Pyreverse

PR #23, opened 2020-05-06
Add complex policies (part 2, move activity tours closer to home)

KasiaKoz:
Adds a modifier: `MoveActivityTourToHomeLocation` class which takes activities list. Things in the list are considered as tours. 

A tour is a chain of activities sandwiched by two home activities (or boundary of persons plan e.g. if person's activities are: `'work'`, `'home'`, `'shop'`, `'work'`, `'home'` (in that order), their tours are: `['work']` and `['shop', work']`). 

If you give this modifier `['shop']` then a tours `== ['shop']` will be moved to home location. The sampling and filtering methods from https://github.com/arup-group/pam/pull/19 holds, i.e. they will be moved if they satisfying filtering conditions and are sampled for being moved.

In the above example, there are no matching tours, meaning the shop activity on the way to work would not be moved closer to home.

A shopping example available for perusal in [Complex policies notebook](https://github.com/arup-group/pam/blob/kasia-add-complex-policies-part-2/notebooks/PAM_Complex_Policies.ipynb)


pam/modify.py
```diff
@@ -63,164 +98,389 @@ def apply_to(self, household):
...
-
-    def evaluate_household_policy(self, household):
-        if self.probability_level == 'household':
-            if self.household_is_selected:
-                self.remove_household_activities(household)
-        elif self.probability_level == 'person':
-            for pid, person in household.people.items():
-                if self.person_satisfies_attribute_conditions(person) and self.person_is_selected(person):
-                        self.remove_household_activities(household)
+    def apply_to(self, household, person=None, activities=None):
```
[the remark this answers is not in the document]
fredshone: no worries - got it - they are pam objects

tests/test_2_activity_plan.py
```diff
@@ -126,3 +126,92 @@ def test_reverse_iter():
...
+    plan = Plan(1)
+    for i in range(len(activities_and_tour['activities'])):
+        plan.add(activities_and_tour['activities'][i])
+        plan.add(Leg(1))
+    other_act = Activity(8, 'other', 'e')
+    plan.add(other_act)
+
+    assert plan[0].act != plan[-1].act
+
+    assert plan.activity_tours() == activities_and_tour['tours'] + [[other_act]]
```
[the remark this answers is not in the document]
KasiaKoz: added additional test
```
def test_activity_tours_segments_home_to_other_act_nonhome_looped_plan(activities_and_tour):
    other_act = Activity(8, 'other', 'e')

    plan = Plan(1)
    plan.add(other_act)
    plan.add(Leg(1))
    for i in range(len(activities_and_tour['activities'])):
        plan.add(activities_and_tour['activities'][i])
        plan.add(Leg(1))
    plan.add(other_act)

    assert plan[0].act == plan[-1].act
    assert plan.activity_tours() == [[other_act]] + activities_and_tour['tours'] + [[other_act]]
```
i.e. the tours are still split by `'home'` activity

tests/test_2_activity_plan.py
```diff
@@ -126,3 +126,92 @@ def test_reverse_iter():
...
+    plan.add(Activity(1, 'home', 'a'))
+    plan.add(Leg(1))
+    plan.add(Activity(2, 'shop', 'b'))
+    plan.add(Leg(2))
+    plan.add(Activity(3, 'home', 'a'))
+
+    plan.move_activity(2)
+
+    assert plan[1].end_location == 'a'
+    assert plan[3].start_location == 'a'
```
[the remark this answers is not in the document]
KasiaKoz: it will only update the leg that exists after or before the activity respectively [link](https://github.com/arup-group/pam/blob/kasia-add-complex-policies-part-2/pam/activity.py#L439)

Added these to test this
```
def test_move_activity_at_start_of_plan_updates_leg():
    plan = Plan('a')
    plan.add(Activity(1, 'shop', 'b'))
    plan.add(Leg(1))
    plan.add(Activity(2, 'home', 'a'))

    plan.move_activity(0)

    assert plan[1].start_location == 'a'


def test_move_activity_at_end_of_plan_updates_leg():
    plan = Plan('a')
    plan.add(Activity(1, 'home', 'a'))
    plan.add(Leg(1))
    plan.add(Activity(2, 'shop', 'b'))

    plan.move_activity(2)

    assert plan[1].end_location == 'a'
```

pam/modify.py
```diff
@@ -5,39 +5,74 @@
 import random
 
 
-class Policy:
+class PersonAttributeFilter:
```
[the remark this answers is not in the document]
KasiaKoz: I agree

tests/test_5_MoveActivity.py
```diff
@@ -18,9 +18,141 @@ def assert_correct_activities(person, ordered_activities_list):
...
+def test_MoveActivityTourToHomeLocation_throws_exception_if_apply_to_given_wrong_input(Bobby):
+    policy = modify.MoveActivityTourToHomeLocation([''])
     with pytest.raises(NotImplementedError) as e:
         policy.apply_to(Bobby)
-    assert 'Watch this space' \
+    assert 'Types passed incorrectly: <class \'pam.core.Person\'>, <class \'NoneType\'>, <class \'NoneType\'>. You need <class \'type\'> at the very least.' \
            in str(e.value)
+
+
+def test_MoveActivityTourToHomeLocation_move_individual_activities_moves_an_activity_for_Bobby(Bobby):
```
mfitz, line 56: I don't think what we're testing here is anything about Bobby specifically. A more generic description of what we're testing in the name would make clearer the intent of the test. There is also an implicit assumption about Bobby in this test - I think we're assuming he has an education activity at a specific place in his plan, right? The test isn't clear about this, and that coupled with the name makes it harder to understand the behaviour we're testing here.

Maybe use some assertions to illustrate what it is about Bobby we care about before we make the call to move one of his activities? And think about a better name for the test?

tests/test_5_MoveActivity.py
```diff
@@ -18,9 +18,141 @@ def assert_correct_activities(person, ordered_activities_list):
...
+    policy = modify.MoveActivityTourToHomeLocation(['education'])
+    policy.move_individual_activities(Bobby, [Bobby.plan[2]])
+
+    assert Bobby.plan[2].location == Bobby.home
+
+
+def test_MoveActivityTourToHomeLocation_doesnt_move_individual_activities_for_Bobbys_empty_selected_activities_list(Bobby):
```
mfitz, line 63: The above comment applies to several of these tests, I think.

tests/test_5_MoveActivity.py
```diff
@@ -18,9 +18,141 @@ def assert_correct_activities(person, ordered_activities_list):
...
+            [Activity(5, 'work', 'c'), Activity(6, 'shop', 'd')]]
+
+@pytest.fixture()
+def plain_filter():
+    def filter(act):
+        return True
+    return filter
+
+
+def test_matching_activity_tours_delegates_to_plan_activity_tours(mocker, plain_filter):
```
mfitz, line 100: About this pattern of test - I know I did something like this in Puma, so it's my fault, but wherever possible with testing, you want to focus on behaviour and avoid coupling your test to implementation detail. You could argue that a method that you call making use of a helper method is its own business, and not something we should care about in a test. What we usually care about ultimately is the *effect* of calling that first method, and not so much *how* that method produces the effect.

Relying on too much knowledge of internal implementation detail can make tests brittle, for example if the helper method you are mocking changes, say is refactored into two or whatever, you would possibly then have to change this test, even if the behaviour (the *what*) hasn't changed at all (only the *how* has changed). Where the helper function is intended to be used only internally, and should not be considered part of the public API of a class or module, consider making it private, especially if the helper is only used in one place.

When we find ourselves mocking one of our own helper functions purely to confirm that it has been called, it's a bit of a smell, I think. No need to change these tests, but something to think about in future.

tests/test_5_policies.py
```diff
@@ -653,3 +662,199 @@ def discrete_sampler(obj, mapping, distribution):
     steve = household.people['1']
 
     modify.RemoveActivity.remove_person_activities.assert_called_once_with(steve)
+
+
+##### MoveActivityToHomeLocation
+
+
+def test_MoveActivityToHomeLocation_moves_shop_to_home_location():
+    Hilda = Person(1, attributes={'age': 45, 'job': 'influencer', 'gender': 'female'})
```
mfitz, line 671: I like having this setup right here inside the test - it adds a lot of important context and makes it clear why we expect the assertion at the end of the test to be true.

mfitz: P.S. `'job': 'influencer'`? 😂

tests/test_5_policies.py
```diff
@@ -653,3 +662,199 @@ def discrete_sampler(obj, mapping, distribution):
...
+    Hilda.add(Activity(5, 'home', 'a', start_time=mtdt(17 * 60), end_time=END_OF_DAY))
+    hhld = instantiate_household_with([Hilda])
+
+    policy = modify.PersonPolicy(
+        modify.MoveActivityTourToHomeLocation(['shop']),
+        modify.PersonProbability(1)
+    )
+    policy.apply_to(hhld)
+
+    assert Hilda.plan[2].location == Hilda.home
```
mfitz, line 685: Could be even more explicit here by comparing the activity we now find in this place to an actual activity, so something like `Activity(2, 'home', 'a', start_time=mtdt(8 * 60 + 30), end_time=mtdt(16 * 60 + 30)))` or whatever.

tests/test_5_MoveActivity.py
KasiaKoz: it has nothing to do with bobby specifically, just decided to use him because i like him, I made the names more generic

PR #24, opened 2020-05-06
Plotting fredits

fredshone:
rough fredits for plotting


fredshone:
Adds a few convenience methods:
`population.random_person()`
`population.random_household()`
`person.plot()`
`household.plot()`

Activity plotting is overhauled (looking fly now) and i added two tests to try and maintain coverage. Not keen to tackle plotting tests in the short term.

Some minor edits to parse module also seem to have triggered a big merge diff in parse and activity modules, but these can be largely ignored.

pam/plot.py
```diff
@@ -1,61 +1,138 @@
...
+def build_cmap(df):
+    colors = plt.cm.Set3.colors[::-1]
     activities_unique = df['act'].unique()
+    d_color = dict(zip(activities_unique, colors))
+    d_color['Travel'] = (.3,.3,.3)
+    return d_color
 
 
-    d_color = dict(zip(activities_unique,colors))
+def plot_activities(df):
```
KasiaKoz, line 55: I love the plots @fredshone . I think we should have three functions down the line though: 
- This `plot_activities` with greyed out legs, 
- `plot_legs` with greyed out activities and 
- `plot_plans` with allll the colours

PR #26, opened 2020-05-07
Add complex policies (part 3 - Reduce shared activities)

KasiaKoz:
Hopefully the last complex policy PR for a little while.

Adds new modifier class `ReduceSharedActivity` which works only with `HouseholdPolicy`.

Given a household, it uses a new `Household` method `shared_activities` which gives a list of `Person`s `Activity`ies that satisfy `Activity.is_exact(other)` (location, type of activity, start and end time the same). It then looks at which activities are actually relevant for removal (as given in `activities` list when instantiating the `ReduceSharedActivity` class). It picks a person that will retain the shared activity at random. 
When it comes to removing the right activities it's a little complicated. I created a method based on `Activity.is_exact(other)` but that's too stringent. For example if we define `ReduceSharedActivity(activities=['shop1', 'shop2'])`, if there is a person where the activities adjoin, i.e.

```
...
6:	Activity(4 act:shop_1, location:b, time:18:20:00 --> 18:50:00, duration:0:30:00)
7:	Leg(4 mode:walk, area:b --> b, time:18:50:00 --> 19:00:00, duration:0:10:00)
8:	Activity(5 act:shop_2, location:b, time:19:00:00 --> 19:50:00, duration:0:50:00)
...
```

the method will delete one of the shopping activities and update the time of adjoining activities. This will make the other activities not satisfy the `is_exact` anymore and so it won't be removed.

In the end I went with `act in shared_activities_for_removal` which relies on `Activity.__eq__()` method which only matches location and type of activity. This means both of the shopping activities will be removed **and** any other activity of the same type and location that individual has. I thought this was less likely to happen and more desirable than not removing the chosen shared activities. Let me know your thoughts.

Thank you for reviewing all of my PRs 😭


tests/fixtures.py
```diff
@@ -18,7 +18,7 @@ def assert_correct_activities(person, ordered_activities_list):
     assert len(person.plan) % 2 == 1
     for i in range(0, len(person.plan), 2):
         assert isinstance(person.plan.day[i], Activity)
-    assert [a.act for a in person.plan.activities] == ordered_activities_list
+    assert [a.act for a in person.plan.activities] == ordered_activities_list, [a.act for a in person.plan.activities]
```
mfitz, line 21: Is this change to add a failure message? If so, the message could be clearer.

tests/test_2_activity.py
```diff
@@ -35,4 +47,23 @@ def test_Activities_end_time_differ_and_are_not_exact():
     a_1 = Activity(1, 'act', 'loc', start_time=mtdt(18 * 60), end_time=mtdt(19 * 60))
     a_2 = Activity(2, 'act', 'loc', start_time=mtdt(18 * 60), end_time=mtdt(19 * 60 +1))
 
-    assert not a_1.is_exact(a_2)
\ No newline at end of file
+    assert not a_1.is_exact(a_2)
+
+
+def test_activity_in_list_exact(list_of_acts):
```
mfitz, line 53: We're straying away from the does-something-when-something style of naming here. The behaviour we're testing seems to be that an activity recognises itself as an exact match in a list in which it is a member. Ideally, the test name should make it clear that is what we're testing. The same comment applies to all the tests below in this same file.

tests/test_5_ReduceSharedActivity.py
```diff
@@ -0,0 +1,185 @@
...
+@pytest.mark.xfail() # fails because of the bug
+def test_remove_activities_removes_relevant_shared_activities(Betty):
+    Betty.plan[6].act = 'shop_1'
+    Betty.plan[6].location = Betty.plan[12].location
+    assert_correct_activities(person=Betty, ordered_activities_list=['home', 'escort', 'work', 'shop_1', 'escort', 'home', 'shop_1', 'shop_2', 'home'])
+    modify.ReduceSharedActivity(['shop_1', 'shop_2']).remove_activities(Betty, shared_activities_for_removal=[Betty.plan[12], Betty.plan[14]])
+    assert_correct_activities(person=Betty, ordered_activities_list=['home', 'escort', 'work', 'shop_1', 'escort', 'home'])
+
+
+def test_remove_household_activities_does_nothing_if_people_in_hhld_dont_share_acts(SmithHousehold):
```
mfitz, line 72: This test name is very clear, and I like it :-)

pam/activity.py
```diff
@@ -709,6 +709,12 @@ def is_exact(self, other):
         return (self.location == other.location) and (self.act == other.act) \
                and (self.start_time == other.start_time) and (self.end_time == other.end_time)
 
+    def in_list_exact(self, activities: list):
```
fredshone, line 712: rename as isin()?

PR #31, opened 2020-05-17
segmented OD matrices

IseulSong:
The current code can extract a segmented OD matrix by mode, occupation, trip start time, or trip purpose, only one at once. Please find my test code too for this. However, I'm not sure how I can make the function to apply multiple conditions at the same time  e.g. a OD by  mode 'car' and purpose 'work' at the same time (assuming that's what we need). Please let me know if you have any suggestions for me.


pam/od_matrices_seg.py
```diff
@@ -0,0 +1,49 @@
...
+def od_seg(population, path, mode_seg=None, type_seg=None, time_seg=None, purp_seg=None):  #time in minutes  
+    ozone = []
+    dzone = []   
+
+    if time_seg:
+        time_seg = mtdt(time_seg)
+
+    for hid, household in population.households.items():
+        for pid, person in household.people.items():
+            for p in person.plan:
```
fredshone, line 15: Person has a legs iterator - 
(here: https://github.com/arup-group/pam/blob/6d00e07f9df6a65afa86c0ad5c28111f6b9754ba/pam/core.py#L151)
so you can directly loop through legs without if statement:
```
for leg in person.legs:
    if leg.mode == mode_seg:
```

pam/od_matrices_seg.py
```diff
@@ -0,0 +1,49 @@
+import pandas as pd
+from datetime import datetime
+from pam.utils import minutes_to_datetime as mtdt
+from pam.variables import END_OF_DAY
+
+def od_seg(population, path, mode_seg=None, type_seg=None, time_seg=None, purp_seg=None):  #time in minutes
```
fredshone, line 6: Good to note that time is in minutes. There are a few good practices for adding comments to functions. The one we tend to use is this:
```
def od_seg(...):
    """
    Function to create an od matrix filtered on mode, type, time or purpose.
    """
```
We definately need to catch up on this with the code in general but the aim is to make it easy when we inevitably pick up the code again in the future. There are some ok examples in the activity module: https://github.com/arup-group/pam/blob/6d00e07f9df6a65afa86c0ad5c28111f6b9754ba/pam/activity.py#L381

pam/test_od_matrices_seg.py
```diff
@@ -0,0 +1,71 @@
+import pandas as pd
```
mfitz, line 1: I **love** to see unit tests - well done!

However, this test needs to be moved to the `/tests` directory, otherwise it will not be found by the test runner and will thus never be run in the code coverage report or the build pipeline. All tests need to be somewhere under `/tests`. If you do refactor the `seg` stuff in to `od_matrices`, then you can cut and paste this test in to `tests/test_od_matrices.py`.

pam/od_matrices_seg.py
```diff
@@ -0,0 +1,49 @@
+import pandas as pd
```
mfitz, line 1: Could all of the functionality in `od_matrices_seg` be moved to the existing `od_matrices` module? That looks like the natural place for it to live, rather than in its own module. Any code that produces OD matrices should probably all be in the same module.

pam/test_od_matrices_seg.py
```diff
@@ -0,0 +1,71 @@
...
+    household.add(person)  
+    population.add(household)
+
+    od_matrix_file = os.path.join(tmpdir, "od_matrix_seg_test.csv")
+
+    od_seg(population, od_matrix_file, mode_seg='car') 
+
+    od_matrix_csv_string = open(od_matrix_file).read()
+
+    assert od_matrix_csv_string =='ozone,Barnet,Ealing,Southwark,"Westminster,City of London"\nBarnet,0,0,1,0\nEaling,0,0,0,1\nSouthwark,1,0,0,0\n"Westminster,City of London",0,1,0,0\n'
```
mfitz, line 71: We should always keep line length <= 120 characters, for readability. To reduce the line length and also make it easier to understand what the expected OD matrix should look like, I would format this assertion like this:

```
    expected_od_matrix = \
        'ozone,Barnet,Ealing,Southwark,"Westminster,City of London"\n' \
        'Barnet,0,0,1,0\n' \
        'Ealing,0,0,0,1\n' \
        'Southwark,1,0,0,0\n' \
        '"Westminster,City of London",0,1,0,0\n'
    assert od_matrix_csv_string == expected_od_matrix
```

pam/od_matrices_seg.py
IseulSong: The code is updated as per the comment

pam/test_od_matrices_seg.py
IseulSong: The code is updated as per the comment

IseulSong:
@fredshone I have tried to cover 1.1 leg attribute and 1.2 person attribute in the updated code (started them as separate functions as you suggested but there were a lot of duplication so I combined them into one in the end). But I wasn't sure what you mean by 'activity' in leg attribute - can you explain it a bit further for me? Afterwards, more than happy to move on further if you think I'm on the right track.

pam/od_matrices_seg.py
IseulSong: Replaced existing `write_od_matrices` function with the one with filtering options.

pam/test_od_matrices_seg.py
IseulSong: Replaced the existing `test_writes_od_matrix_to_expected_file` in the `test_8_write.py` with a new test code including segmentation.

pam/od_matrices_seg.py
IseulSong: The code is updated as per the comment

IseulSong:
@fredshone @mfitz some more feedback on the updated code would be very helpful!

pam/write.py
```diff
@@ -43,38 +44,73 @@ def write_travel_diary(population, path, attributes_path=None):
...
+                            'Start time': leg.start_time,
+                            'End time': leg.end_time}                
+                if person_filter:
+                    legs.append({**data, **person.attributes})
+                else:
+                    legs.append(data)         
+                                
+    data_legs = pd.DataFrame(data=legs)
+    
+    if leg_filter or person_filter:
```
fredshone, line 88: I think better these be separate so that it is clearer that it is one OR the other, can also add a return after each one so that the method ends, ie:
```
if leg_filter:
   ...
  return None
elif person_filter:
  ...
  return None
elif time_minutes_filter:
  ...
  return None
```
Otherwise great use of pandas - this approach will extend well.

IseulSong:
Changes have been made according to the comments.

fredshone:
@IseulSong This is ready to merge... Time to press the merge button!

PR #33, opened 2020-05-18
Simplify Policies

KasiaKoz:
This PR doesn't add much new functionality at all but the usage around defining and applying policies is changing. I'm keen to get feedback on both the code and whether you think this makes `pam` easier to use - where can I do better?

- Splits pam.modify into
    - pam.policy.filters:
       - with base class `Filter` 
classes_Pyreverse
    - pam.policy.modifiers
       - with base class `Modifier`, before these classes were subclass of Policy, now they are similar but not subclass of the `Policy` base class 
classes_Pyreverse
    - pam.policy.policies
       - with base classes `Policy` and `PolicyLevels` 
classes_Pyreverse
    - pam.policy.probability_samplers
       - with base class `SamplingProbability` 
classes_Pyreverse

- Adds pre-baked policies (updated this [notebook](https://github.com/arup-group/pam/blob/kasia-spring-clean-policies/notebooks/PAM_Simple_Policies.ipynb) to use these):
    - `RemoveHouseholdActivities`: Household level policy that removes activities
    - `RemovePersonActivities`: Person level policy that removes activities
    - `RemoveIndividualActivities`: Activity level policy that removes activities
    - `MovePersonActivitiesToHome`: Person level policy that moves activities to home location
    - `ReduceSharedHouseholdActivities`: Household level policy that removes shared activities within a household from all but one random person in the household.

- Updates `apply_policies` to work as both modifying current `Population` objects or returning a modified copy by specifying `in_place=bool`

- Usage: 
    - All the above classes can be imported using `from pam.policies import X` Screenshot 2020-05-18 at 14 26 28
    - Added docstrings Screenshot 2020-05-18 at 14 27 39
    - All classes in `pam.policy` have descriptive printing method and representation:
Screenshot 2020-05-18 at 17 17 52

- Moves a lot of checks that were repetitive across same level classes to the superclasses. 
    - e.g. checks for `0<p<=1` is in `SamplingProbability`
    - e.g. checking for `modifier` to be of `Modifier` class and `attribute_filter` to be of `Filter` class is done in `PolicyLevel` class
- Improves typing (suggests types of variables) which is a suggestion in python, but added checks for types and hopefully more descriptive error messages

Accidental Bonus:
- Adds kwargs to plots for consistent cmap (generated separately using the functions in plot module) or saving to file for higher resolution 

Some other things I wanted to tackle but have left out for now:
- Pre bake some callables to use when defining `conditions` for `Filter`s 
- Make using a callable within probability samplers easier and probably pre bake some stuff for that too.


pam/policy/filters.py
```diff
@@ -0,0 +1,88 @@
...
+            for attribute_key, attribute_condition in self.conditions.items():
+                satisfies_attribute_conditions &= attribute_condition(person.attributes[attribute_key])
+            return satisfies_attribute_conditions
+        elif self.how == 'any':
+            satisfies_attribute_conditions = False
+            for attribute_key, attribute_condition in self.conditions.items():
+                satisfies_attribute_conditions |= attribute_condition(person.attributes[attribute_key])
+            return satisfies_attribute_conditions
+        else:
+            raise NotImplementedError('{} not implemented, use only `all` or `any`'.format(self.how))
```
mfitz, line 88: Maybe this validation should live in the constructor?

pam/policy/modifiers.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def apply_to(self, household, person=None, activities=None):
+        if (activities is not None) and (person is not None):
+            self.remove_individual_activities(person, activities)
+        elif person is not None:
+            self.remove_person_activities(person)
+        elif household is not None and isinstance(household, pam.core.Household):
+            self.remove_household_activities(household)
+        else:
+            raise NotImplementedError('Types passed incorrectly: {}, {}, {}. You need {} at the very least.'
```
mfitz, line 65: Maybe `TypeError` is a better fit when the type is wrong, and some other error for when `None` is semantically wrong?

pam/policy/modifiers.py
```diff
@@ -0,0 +1,270 @@
...
+    List of activities that should be considered for sharing. Does not
+    require an exact match. E.g. if passed ['shop_food', 'shop_other']
+    if a household has an only 'shop_food' shared activity, that will
+    be reduced.
+    """
+    def __init__(self, activities: List[str]):
+        super().__init__()
+        self.activities = activities
+
+    def apply_to(self, household, person=None, activities=None):
```
mfitz, line 135: Is there a reason for not using type hints here?

pam/policy/modifiers.py
```diff
@@ -0,0 +1,270 @@
...
+        if household is not None and isinstance(household, pam.core.Household):
```
mfitz, line 136: I think `isinstance` against `None` will evaluate to false, so perhaps you can lose `None` check?

pam/policy/modifiers.py
```diff
@@ -0,0 +1,270 @@
...
+                    if person != person_retaining_activities:
+                        self.remove_activities(person, acts_for_removal)
+
+    def is_activity_for_removal(self, p):
+        return p.act.lower() in self.activities
+
+    def shared_activities_for_removal(self, household):
+        shared_activities = household.shared_activities()
+        shared_activities_for_removal = []
+        for act in shared_activities:
```
mfitz, line 176: Could you use a list comprehension here? Arguably more "pythonic".

pam/policy/probability_samplers.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def p(self, x):
+        if isinstance(x, pam.core.Household):
+            return self.compute_probability_for_household(x)
+        elif isinstance(x, pam.core.Person):
+            raise NotImplementedError
+        elif isinstance(x, pam.activity.Activity):
+            raise NotImplementedError
+        else:
+            raise NotImplementedError
```
mfitz, line 89: TypeError for the final else, maybe?

tests/test_5_filters.py
```diff
@@ -1,41 +1,53 @@
...
+    attrib_filter = PersonAttributeFilter({})
+    assert '{}'.format(attrib_filter.__class__.__name__) in attrib_filter.__repr__()
+
+
+def test_subclass_name_features_in_str_string():
+    attrib_filter = PersonAttributeFilter({})
+    assert '{}'.format(attrib_filter.__class__.__name__) in attrib_filter.__str__()
 
 
 def test_PersonAttributeFilter_satisfies_conditions_delegates_to_household_satisfies_conditions_when_given_household(mocker):
```
mfitz, line 23: Not keen on the camel case in test names.

tests/test_5_modifiers.py
```diff
@@ -0,0 +1,20 @@
+from tests.fixtures import Bobby
```
mfitz, line 1: Do we still need the `_5_` in the module name?

KasiaKoz: dunno, not keen on those numbers in general but I went with the flow

PR #37, opened 2020-06-08
Geo sampling xml output

arup-sb:
Adds 
- geometry (random) sampling object (basically pulled from Mimi), can be replicated with @fredshone's facility later 
- write_matsim_xml now with shiny x, y coordinates for Matsim plans.xml

backwards compatible, so will only try to do location sampling if include_loc is passed to load_travel_diary(...)

There is a notebook to demo the functionality at notebooks/pam-geo-sampling.ipynb


arup-sb:
> I have some duplicate functionality for random spatial sampling which i think is a bit better than what i did in mimi so would prefer to use that. I will PR shortly. Sorry for this confusion. I got pulled into TII shenanigan so didn't finish - i will work on this pm.

Seems like duplicated effort? The introduced above method is functionally valid and tested already. Would be interested if you have specific feedback on the method though.

> Broadly we'd (PAM incorporated) like to support lots of different inputs (the existing read methods are for matsim and diary formats aka lopops). We could additionally suport a simple tour based input or even od matrices.
> 
> At the moment i am more in favour of adding additional read methods than extending existing ones.

The method extension allows reading of location data which matsim plans need. The previous method was in a sense not functionally complete. Also, the extension is backwards compatible. I don't really take your point here.

> Regarding spatial sampling - I think this should be done via the PAM api rather than pre. Because all inputs will likelly be area based and require some form of location sampling. Rather than trying to suuport all these in the read module i prefer to more generically use the API, ie loop through plans after they are loaded.

This PR introduces a sampling class object that could be used inside of the PAM api or pre. I dont think there is really the constraint that you are describing. That being said, why would sampling internally be exclusively preferred? You ruin data provenance if say you wanted to read in data with pre-defined locations?

fredshone:
The samplers and their pattern have been planned for a while. Nothing wrong with the functionality of your code and example but i would prefer to encourage spatial sampling after loading a dataset rather than suport every case at read.

My implementation of the random point sampler is also tested and i made some improvements (i think) over the mimi implementation. So I'd like to at least consider using that instead. 

>This PR introduces a sampling class object that could be used inside of the PAM api or pre. I dont think there is really the constraint that you are describing. That being said, why would sampling internally be exclusively preferred? You ruin data provenance if say you wanted to read in data with pre-defined locations?

I agree. It is not a constraint. My concern is more about interface than functionality. 9 times out of 10 we are expecting to load area based locations from tabular data. So it seems to me that we can more simply support spatial/facility/other sampling in the core format rather than for each read method.

Your point about provenance is good but in all cases we will be writing to matsim which will preserve locs. If this changes then agree we should consider new in/out formats. But i do not wish to artificially add some now.

arup-sb:
> The samplers and their pattern have been planned for a while. Nothing wrong with the functionality of your code and example but i would prefer to encourage spatial sampling after loading a dataset rather than suport every case at read.

The sampling class can be used for downstream sampling, so I don't really take this point. It's not handling every case at read, its handling the reading of location (point) inputs, which is an important use case, and probably essential going forward for things like existing matsim plans, gps data.

> My implementation of the random point sampler is also tested and i made some improvements (i think) over the mimi implementation. So I'd like to at least consider using that instead.

I'm definitely up to consider it so please share some specifics. I would rather use mine is not really a valid argument.

> I agree. It is not a constraint. My concern is more about interface than functionality. 9 times out of 10 we are expecting to load area based locations from tabular data. So it seems to me that we can more simply support spatial/facility/other sampling in the core format rather than for each read method.

The interface is backwards compatible and the Sampling classes can be used internally for sampling in the core format case so this point doesn't really make sense.

> Your point about provenance is good but in all cases we will be writing to matsim which will preserve locs. If this changes then agree we should consider new in/out formats. But i do not wish to artificially add some now.

Again, handling location data probably should have been done from the start. So the fact that it wasn't implemented yet isn't a good reason for not integrating it now.

PR #44, opened 2020-08-05
Dft read method

fredshone:
FYI

Code and notebook added for prelim England Population (National Travel Survey 2016 & 17).

Looks to be in a reasonable state for some England scenarios... there are about 140k unique diary days which seems ok for most country scale work. 

Still WIP (needs tests, weighting and spatial sampling).

But csvs for the 1% (unweighted) sample here: https://arup.sharepoint.com/:f:/s/TransportAnalytics/Egi4kEtkZflMkiKPO3aLlm8B2bB2l6ndJjJNRjSNJ_RFLw?e=PVcmBH

Note all the available person and household attributes described in the .rtf files (Iseul you have already seen these). There are a lot of interesting attributes available for hhs, people and even vehicles and so on.

Next steps for this PR:
- test
- weighting
- geographic mapping (the Unitary Authority codes they use do not match and geographies i've tried)
- facility sampling (there are some unusual tech challenges around this due to large area sizes)


elladahan:
Looks cool! 
Also a few questions from my side:
1. I saw that you mentioned trip purpose is slightly different to how usually we process it. Does that mean we need to add logic to convert those fields into what we would expect in trip purpose? 
2. Are there any other problems with mapping the DfT data to the available PAM fields? 
2. Are there any changes needed for the csv exports which are used for the dashboard? Would we want to add those additional fields on household, person, etc?

fredshone:
Thanks guys - good questions:

1. `How are you accessing the DfT raw csv files, trip_csv, households_csv, individuals_csv? Do they need to be specifically queried to a database or something? (they don't seem to be in the data that you've uploaded in the sharepoint!)`
These are included in the UK Data Service data (I just haven't shared) - I have them downloaded locally. It would be great if you'd both get accounts with the UK Data Service then join our 'project' via their portal (PN:193378 "UK Population Synthesis"). I'm not sure exactly how this process works though.

2. `UA codes seem to be very wide geographical areas - some of the journey purposes are likely to be local trips like shopping, education. Would it make a bit difficult to see the impacts of certain policies on trip demands?`
Note that I still havn't found a geography that matches the ids or the names. But it is described and looks a lot like UAs. The high aggregation is a concern, but we can still see that a trip took place. The difficulty will be that we have very poor information about distance. But we do have good info on trip mode and duration which can be used as a proxy. Related to this - the next step for me might be to do facility sampling - which will map the activity locations to coordinates. I think the existing facility sampling methodology is too dumb with this coarse geography so I am thinking to do something smarter that takes into account distance between sampled activity points...

3. `Is there a specific reason that we are sampling 1% of the data? Then why are we doing weighting/ expansion of trips again after sampling them?`
Yes - I was in a terrible rush. I will progress to a representative (ie weighted) English households sample soon. I provide this little prelim population as a proof of concept and dummy data to start progressing any analysis stuff.

1. `I saw that you mentioned trip purpose is slightly different to how usually we process it. Does that mean we need to add logic to convert those fields into what we would expect in trip purpose?`
It was a new type of diary format - with each trip having a from and to activity classification. This was actually more info that we normally get. In any case it is now in the regular PAM format so everything will work as before. There are new activity descriptions, which i will probably simplify in the next iteration.

2. `Are there any other problems with mapping the DfT data to the available PAM fields?`
It works ok. There are loads of new attributes for people and households but these are stored as generic python dictionaries so doesn't cause an issue. I only include a small subset of the available attributes, to speed processing and not cripple memory (if you check out the .rtf files you can see there a 100s of fields). I am currently ignoring data about vehicles etc. But this could be incorporated.

3. `Are there any changes needed for the csv exports which are used for the dashboard? Would we want to add those additional fields on household, person, etc?`
No - this should all work as before, except that there are far more categories of mode and activity which will makle the charts look busy. But i will simplify these down to a more sensible number of core acts and modes next iteration.

fredshone:
Hi all, this PR is updated. The notebook is somewhat complete and renamed `NTS Population Synthesis Example`. Where NTS is the National Travel Survey. I still keep the data separate from the project, but if you want to play then you need a UK Data Service loggin (https://beta.ukdataservice.ac.uk) and request to join or 'project' (PN 193378).

1. The notebook has more notes and description of what is going on
2. I have not solved the geometry encoding, instead I am just showing diaries that are entirelly within either Inner or Outer London. So this is a pretty brutal subset.
3. As Iseul previously commented because of the large geometries, there is horendous lack of realistic variance in activity locations and trips. This is described in more detail below.
4. The sample size is very small.
5. The sample weighting is based on the NTS household weighting. I havn't investigated what this is based on or even what the unit is.
6. For Arup people, the outputs are available on sharepoint. You will be able to load the xml into pam if you want to have a play using ~`pam.read.read_matsim()`

Two Big critiques:

**1. Poor Facility Sampling**
The large areas used for the original diary data are pretty bad for facility sampling (Inner/Outer London). For example we fail to capture realistic work place density in Canary Wharf:

kepler-gl

**2. Poor origin/destination sampling**
Similarly the large areas create a lot of nuts trips. For example super long walks or shopping trips to the other side of London:

kepler-gl(1)

In general this would be greatly reduced by access to the more precise datasets (which tend to use LSOAs for example). However there is also the oportunity to improve the facility sampler so that it makes better choices given known trip attributes such as mode choice and duration.

PR #47, opened 2020-10-07
Plot trips spatially

KasiaKoz:
This PR adds:

- reading service ids, route ids and network routes from agents plans in read_matsim method. Adds those as optional attributes for the Leg class.
- methods for Population, Household and Person to build a geopandas.GeoDataFrame for all the Legs, there are columns for pid and hid to differentiate between people. The Geometry column is a LineString with two points (because no network to get the actual route, but maybe when coupled with GeNet....)
- methods to plot these GeoDataFrames with plotly, given a mapbox token
- a jupyter notebook that shows what you can do, also how to save the geo dataframe to geojson. I had to scrap the actual plotly visualisation because the meta data had my api key in it, so you _should_ just be able to see png pictures of the plots


fredshone (approved):
nice nice nice

gac55 (approved):
Excellent.

PR #48, opened 2020-11-10
Writing matsim XML's - Return the object, not the key in case of a dict

gac55:
An issue with expected types when writing to Matsim xml.

Output before fix:
```
  <object id="10002061023">
    <attribute class="java.lang.String" name="psex">{10002061023: 'Male'}</attribute>
    <attribute class="java.lang.String" name="page">{10002061023: 40}</attribute>
    <attribute class="java.lang.String" name="pdlicense">{10002061023: 1}</attribute>
    <attribute class="java.lang.String" name="hincome">{10002061023: 4}</attribute>
    <attribute class="java.lang.String" name="hsize">{10002061023: 4}</attribute>
    <attribute class="java.lang.String" name="hid">{10002061023: 640586395095}</attribute>
    <attribute class="java.lang.String" name="subpopulation">{10002061023: 'License'}</attribute>
    <attribute class="java.lang.String" name="source">{10002061023: 'phase-2-tfl'}</attribute>
  </object>
```

Output after fix:
```
<object id="10002061023">
    <attribute class="java.lang.String" name="psex">Male</attribute>
    <attribute class="java.lang.String" name="page">40</attribute>
    <attribute class="java.lang.String" name="pdlicense">1</attribute>
    <attribute class="java.lang.String" name="hincome">4</attribute>
    <attribute class="java.lang.String" name="hsize">4</attribute>
    <attribute class="java.lang.String" name="hid">10002061023</attribute>
    <attribute class="java.lang.String" name="subpopulation">License</attribute>
    <attribute class="java.lang.String" name="source">phase-2-tfl</attribute>
  </object>
```
I don't think this is solved, some more effort needs to go into ensuring types are correct. This should happen further upstream, likely at ingestion. At present, users will likely use the types as inferred by Pandas when reading from csv - as was the case with the example above. 

I present this improvement as an incremental change which should at least make debugging better.

### DIFF
--- a/pam/write.py
+++ b/pam/write.py
@@ -204,7 +204,7 @@ def write_matsim_attributes(population, location, comment=None, household_key=No
 
             for k, v in attributes.items():
                 attribute_xml = et.SubElement(person_xml, 'attribute', {'class': 'java.lang.String', 'name': str(k)})
-                attribute_xml.text = str(v)
+                attribute_xml.text = str(v[1])
 
     write_xml(attributes_xml, location, matsim_DOCTYPE='objectAttributes', matsim_filename='objectattributes_v1')
 
--- a/pam/write.py
+++ b/pam/write.py
@@ -209,17 +209,11 @@ def write_matsim_attributes(population, location, comment=None, household_key=No
                 # We should ensure types further upstream
                 # This solves TfL Phase 2 issue 
 
-                if isinstance(v, (str,int)):
+                if not isinstance(v, dict):
                     attribute_xml.text = str(v)
-                
-                elif isinstance(v, dict):
+                else
                     attribute_xml.text = str(list(v.values())[0])
 
-                else:
-                    print(type(v))
-                    print(attributes)
-                    print(pid)
-                    print(v)
 
     write_xml(attributes_xml, location, matsim_DOCTYPE='objectAttributes', matsim_filename='objectattributes_v1')
 
--- a/pam/write.py
+++ b/pam/write.py
@@ -211,7 +211,7 @@ def write_matsim_attributes(population, location, comment=None, household_key=No
 
                 if not isinstance(v, dict):
                     attribute_xml.text = str(v)
-                else
+                else:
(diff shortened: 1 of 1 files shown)


fredshone:
I don't really like this. Prefer that we make it clearer that everything is forced to string. Which i think is what is roughly expected with writing to xml.

Interested to know how you got dictionaries as attributes? Seems there must have been an issue with the read method?

pam/write.py
```diff
@@ -204,7 +204,16 @@ def write_matsim_attributes(population, location, comment=None, household_key=No
...
-                attribute_xml.text = str(v)
+
+                # This is hacky
+                # We should ensure types further upstream
+                # This solves TfL Phase 2 issue 
+
+                if not isinstance(v, dict):
+                    attribute_xml.text = str(v)
+                else:
+                    attribute_xml.text = str(list(v.values())[0])
```
mfitz, line 215: I'm a bit surprised this did not cause a test somewhere to break. Presumably that's because we are not using the kind of data in any of the unit tests that would reproduce the bug? With any bug, we should really start with one or more new tests that fail because of the bug and turn green after we've applied the fix, so I think we need at least one new test.

gac55:
yes, something in the read method, but I can't trace. Will continue to explore

pam/write.py
mfitz: Screen Shot 2020-11-10 at 15 24 18

Couple of gaps identified by the coverage report that we should plug with new tests, I think, especially since one of them (line 215) is at the heart of the change in this PR.

mfitz (changes requested):

gac55:
I am unable to create my bug without using the large input Collins gave me. I don't think it's at read anymore, but somewhere in the middle of the pipeline

gac55:
This issues was caused by duplicate household rows. This resulted in https://github.com/arup-group/pam/blob/master/pam/read.py#L84 returning a dictionary with nested dictionaries. 

I will close this and open an issue.

PR #52, opened 2020-11-11
make activity loader use first activity type, not assume home

gac55:
This makes the `load_activity_plan` method consistent with other methods, where it is not assumed the first activity is a `home` type. 

Came across this when using activity_plan type with freight, delivery plans.

### DIFF
--- a/pam/read.py
+++ b/pam/read.py
@@ -304,10 +304,14 @@ def load_activity_plan(
                 # home_area=home_area
             )
 
+            first_act = trips.iloc[0].activity.lower()
+            if not first_act == "home":
+                logger.warning(f" Person pid:{pid} hid:{hid} plan does not start with 'home' activity")
+
             person.add(
                 activity.Activity(
                     seq=0,
-                    act='home',
+                    act=first_act,
                     area=origin_area,
                     start_time=utils.minutes_to_datetime(0),
                 )


gac55:
I will make a change to the test

gac55:
I am confused by these tests. As an example, consider the [inputted activity plans](https://github.com/arup-group/pam/blob/master/tests/test_data/test_activity_plans.csv) for the 
You can see agent `pid=0` has the following activities:
```
pid | hid | seq | hzone | ozone | dzone | activity | mode | tst | tet | freq
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
  | 0 | 0 | 1 | a | a | b | work | car | 500 | 530 | 10
  | 0 | 0 | 2 | a | b | a | home | car | 1000 | 1030 | 10
```
i.e. `work-home` but the test is looking to assert `home-work-home` [here](https://github.com/arup-group/pam/blob/master/tests/test_3_parse_activity_plan.py#L34)

My change stops the force change of the first activity to `home`, resulting in a failed test. Have I misunderstood the test here @fredshone or @mfitz ?

I can change the test to assert `work-home` or change the inputted test csv if I have understood correctly.

fredshone:
There is some work do be done in clarifying the various read methods.

I think the one you are using was explicitly added for the 'new' (April) TFL data which were home based tours. From the `load_activity_plan` docstring:

```
 Turn Activity Plan tabular data inputs (derived from travel survey and attributes) into core population
    format. This is a variation of the standard load_travel_diary() method because it does not require
    activity inference. However all plans are expected to be tour based, so assumed to start and end at home.
    We expect broadly the same data schema except rather than trip 'purpose' we use trips 'activity'.
```
In which case the test makes sense I think as the first activity is `home` and was required to be inferred. If this is not the case for the data be read then the read method can be adjusted. I would propose adding an option, eg `force_home_base = True`.

gac55:
Ok - understand, makes sense. I will implement a flag as you suggest.

fredshone (changes requested):
see comment, prefer adding a flag as suggested.

gac55:
Raised as an issue [here](https://github.com/arup-group/pam/issues/61)

PR #53, opened 2020-11-13
adds requirements to setup and "removes" some xfails

fredshone:
Thought i found a plan plotting bug but turned out to be my plotly install being weird.

But a few little changes:
- very minor plotting fix for single axis plots
- i've added requirements to the setup
- cleaned up the parse xfails - i leave as comments for my future benefit

@KasiaKoz you have a few xfails and a comment referring to a bug...


pam/plot/plans.py
```diff
@@ -36,21 +36,23 @@ def build_person_df(person):
     """
     Loop through a persons plan, creating a pandas dataframe defining activities for plotting.
     """
-    activities, modes, start_times, end_times, durations = [], [], [], [], []
-
+    data = {
```
mfitz, line 39: This is definitely nicer.

tests/test_3_parse_challenge.py
```diff
@@ -104,36 +99,38 @@ def test_agent_pid_13_not_start_and_return_home_night_worker_complex_chain_type2
...
 
 
-@pytest.mark.xfail()
-def test_agent_pid_14_not_start_from_home_impossible_chain_type1_intra_trip(
-        test_trips, test_attributes
-):
-    population = load_travel_diary(test_trips, test_attributes)
-    acts = [a.act for a in population.households['11'].people['14'].activities]
-    assert acts == ['work', 'shop', 'home']
+# Redundant test - we rarely see this requirment any more
```
mfitz, line 102: I think we should either delete these tests completely, or if we want to keep them and not run them for some reason, use the pytest `skip` decorator (https://docs.pytest.org/en/reorganize-docs/new-docs/user/skipping.html#unconditional-skip). Having them commented out seems like the wrong thing.

mfitz (changes requested):

tests/test_3_parse_challenge.py
fredshone: added
thanks for the link

mfitz (approved):

PR #57, opened 2020-12-03
adds operations

fredshone:
place holder PR, tests still to be added


fredshone:
merging while no one is looking

PR #58, opened 2020-12-09
Benchmarking

Theodore-Chatziioannou:
See tickets LAB-823 and LAB-815 for description. Adds a method for extracting benchmark data from a PAM population, for specified combinations of data fields, dimensions, and aggregation functions.

### DIFF
--- a/pam/write.py
+++ b/pam/write.py
@@ -456,16 +456,16 @@ def write_benchmarks(
     ## aggregate across specified dimensions
     if dimensions != None:
         if data_fields != None:
-            df = df.groupby(dimensions)[data_fields].agg(aggfunc)#.reset_index()
+            df = df.groupby(dimensions)[data_fields].agg(aggfunc)
         else:
-            df = df.value_counts(dimensions)#.reset_index()
+            df = df.value_counts(dimensions)
             
     ## show as percentages
     if normalise_by != None:
         if normalise_by == 'total':
             df = df / df.sum(axis = 0)
         else:
-            df = df.groupby(level = normalise_by).transform(lambda x: x / x.sum())#.sort_index(level = normalise_by)        
+            df = df.groupby(level = normalise_by).transform(lambda x: x / x.sum())     
     df = df.sort_index().reset_index()
     
     ## flatten column MultiIndex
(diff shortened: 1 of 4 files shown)


fredshone (commented):
I like the code and I've had a look at the examples. 

2 comments:

1. Before building something into elara I think we should choose at least one benchmark to take forward (for now) and add a wrapper, eg:
```
def write_mode_distance_benchmark(population):
    return write_benchmarks(population, etc...)

def write_mode_duration_benchmark(population):
    return write_benchmarks(population, etc...)
```

2. Chris Bruce is really keen to bin by trip durations and distances. Can you add this? basically just define some standard bins (in m or km for distance and preferably seconds for duration), eg:

```
def write_mode_distance_bins_benchmark(population):
    return write_benchmarks(population, etc...)
```

Theodore-Chatziioannou:
@fredshone I have added trips by distance and duration bin benchmarks, as well as some other wrappers (hour and mode+purpose breakdowns)

pam/write.py
```diff
@@ -403,3 +403,124 @@ def write_population_csv(list_of_populations, export_path):
...
+                    'ozone': leg.start_location.area,
+                    'dzone': leg.end_location.area,
+                    'seq': seq,
+                    'purp': leg.purp,
+                    'mode': leg.mode,
+                    'tst': leg.start_time.time(),
+                    'tet': leg.end_time.time(),
+                    'duration': leg.duration / pd.Timedelta(minutes = 1), #duration in minutes
+                    # calculate leg euclidean distance in km:
+                    'euclidean_distance': ((leg.end_location.loc.x-leg.start_location.loc.x)**2 + (leg.end_location.loc.y-leg.start_location.loc.y)**2)**0.5 / 1000,
```
fredshone, line 446: small favour - while you have this open, can you make this a Leg method, ie `Leg.euclidean_distance()`.

pam/write.py
```diff
@@ -403,3 +403,124 @@ def write_population_csv(list_of_populations, export_path):
...
+        
+    ## export or return dataframe
+    if path != None:
+        if path.lower().endswith('.csv'):
+            df.to_csv(path, index=False)
+        elif path.lower().endswith('.json'):
+            df.to_json(path, orient='records')
+        else:
+            raise ValueError('Please specify a valid csv or json file path.')
+    else:
```
fredshone, line 499: maybe always return the df? I don't see the harm.

fredshone (approved):
Looks great - I've added two small optional favours/requests.

pam/write.py
Theodore-Chatziioannou: done (as property)

pam/write.py
Theodore-Chatziioannou: done

PR #65, opened 2021-02-09
Facility sampler complex

Theodore-Chatziioannou:
Introduces some of the ideas we have discussed around enabling more advanced facility sampling:
* add a population.sample_locs_complex method which passes additional information to the facility sampler, such as the duration and mode of arriving trip and the previous activity location. This information can be used by the sampler to perform:
* distance-weighted sampling (trying to better align locations to the respective sampled trip durations), and
* enable distance-to-nearest-transit-stop cutoff restrictions

There are also some slight structure changes that allow the new additions to work alongside the previously introduced methods (area-weighted sampling, or simple sampling without replacement), and facilitate future extensions. 

I am also introducing the idea of long-term location choices (such as work and school) which can be allocated only once for a zone, vs short-term, which don't have to be persistent.

Some documentation, caveats and examples can be found in [this ](https://github.com/arup-group/pam/blob/facility_sampler_complex/notebooks/weighted-facility-sampling_demo.ipynb) notebook. 

Overall, I believe the added flexibility and methods can be useful to us, but I am not 100% keen on the distance-based sampler, largely for the reasons I mentioned in our last sprint (will often need calibration, it's hard to "close" the daily path, working with euclidean distances, etc). That said, we can test whether it improves model runtimes and validation, and it can be a basis for further development.


pam/core.py
```diff
@@ -317,6 +321,66 @@ def sample_locs(self, sampler):
...
+        for _, household in self.households.items():
+            home_loc = activity.Location(
+                area=household.location.area,
+                loc=sampler.sample(household.location.area, 'home', previous_mode = None, previous_duration=None, previous_loc = None)
+            )
+            previous_mode = None
+
+            unique_locations = {(household.location.area, 'home'): home_loc}
+
+            for _, person in household.people.items():
```
KasiaKoz, line 339: you're using `_` twice under the same loop, you're not using it but it's best to be safe

pam/variables.py
```diff
@@ -4,3 +4,18 @@
...
+###### FACILITY SAMPLING VARIABLES #########################################################
+EXPECTED_EUCLIDEAN_SPEEDS = {
+    'average':10,
+    'car':20,
+    'bus':10, 
+    'walk':5, 
+    'cycle':15
+} # mode speeds expressed as *euclidean* kilometers per hour 
+
+TRANSIT_MODES = ['bus','rail','pt'] # modes for which the maximum walk distance applies
```
KasiaKoz, line 18: i'd maybe add subway too

pam/variables.py
```diff
@@ -4,3 +4,18 @@
...
 END_OF_DAY = datetime(year=1900, month=1, day=2, hour=0, minute=0, second=0)
+
+
+###### FACILITY SAMPLING VARIABLES #########################################################
+EXPECTED_EUCLIDEAN_SPEEDS = {
+    'average':10,
+    'car':20,
+    'bus':10, 
+    'walk':5, 
+    'cycle':15
```
KasiaKoz, line 15: i notice rail and pt are missing, is that because they should default to average?

KasiaKoz (approved):
I see in the notebook you're already using a math environment $$, you can use `\frac{weight_j}{expected\_distance_{ij}^2}` for `weight_j / (expected\_distance_{ij}^2) ` I think it will make it clearer

good stuff, I appreciate the comments you leave throughout

pam/samplers/facility.py
```diff
@@ -162,19 +180,89 @@ def write_facilities_xml(self, path, comment=None):
...
+                expected_distance = expected_distance * 1000 # convert km to meters
+                distance_weights = np.abs(distances - expected_distance)
+                distance_weights = np.where(distance_weights==0, variables.SMALL_VALUE, distance_weights) # avoid zero weights
+
+                ## normalise weights by distance            
+                # weights = weights.values / np.exp(distance_weights) # exponentiate distances to reduce the effect very small distances 
+                # weights = weights.values / np.exp(distance_weights/distance_weights.max())          
+                # weights = weights / (1 + np.exp(1*(distance_weights - distance_weights.mean()))) # alternative formulation: logistic curve
+                # weights = weights.values / (1 + np.exp((distance_weights))) # alternative formulation: logistic curve   
+                weights = weights.values / (distance_weights ** 2) # distance decay factor of 2
```
fredshone, line 263: This is where the apples and oranges is happening (comparing difference in floor area (m2) to difference in distance (m)).
What do you think about 'standardising' by the maximum then multiplying, eg:
```
floor_area_weights = floor_areas / max_floor_area
trip_distance_weights = (max_distance + variables.SMALL_VALUE - distance_weights) / max_distance
weights = floor_area_weights * trip_distance_weights
```
This is not necesarily better but is easier to explain. ie the weight of the largest floor area is equal to the weight of the minimum distance discrepancy?

fredshone (commented):
Looks good - some comments on variable naming to make it easier to read.

Overall I'd be happy to see these samplers move into their own module and further completely separate the prev simple and new complex methods. I think think the dupliocate code will be worth the clarity. I think there will be further iterations. we can consolidate into a preffered `population.facility_sample()` method once we're confident about weights and such.

Nice work!

Theodore-Chatziioannou:
> I see in the notebook you're already using a math environment $$, you can use `\frac{weight_j}{expected\_distance_{ij}^2}` for `weight_j / (expected\_distance_{ij}^2) ` I think it will make it clearer
> 
> good stuff, I appreciate the comments you leave throughout

thanks - updated formula

pam/core.py
Theodore-Chatziioannou: updated

pam/core.py
```diff
@@ -317,6 +321,66 @@ def sample_locs(self, sampler):
...
+                previous_mode = None
+                previous_duration = None
+                previous_loc = None
+                
+                for idx, plan in enumerate(person.plan):
```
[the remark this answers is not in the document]
Theodore-Chatziioannou: updated to component

pam/samplers/facility.py
```diff
@@ -57,19 +61,24 @@ def __init__(
     def clear(self):
         self.facilities = {}
 
-    def sample(self, location_idx, activity):
+    def sample(self, location_idx, activity, previous_mode=None, previous_duration=None, previous_loc=None):
```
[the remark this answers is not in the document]
Theodore-Chatziioannou: done. I usually try to avoid using `mode` as a variable name in python, but I see what you mean

pam/samplers/facility.py
```diff
@@ -127,8 +139,14 @@ def build_facilities_sampler(self, facilities, zones, weight_on = None):
                 facs = zone_facs.loc[zone_facs.activity == act]
                 if not facs.empty:
                     points = [(i, g) for i, g in facs.geometry.items()]
-                    weights = facs[weight_on] / facs[weight_on].sum() if weight_on != None else None # sum of weights/probabilities should be one
-                    sampler_dict[zone][act] = inf_yielder(points, weights)
+                    if weight_on != None:
```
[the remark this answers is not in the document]
Theodore-Chatziioannou: updated

pam/samplers/facility.py
Theodore-Chatziioannou: @fredshone  The way I would communicate this step it is as weight adjustments/differences:  when choosing between two facilities with the same floorspace, the one being half as far from the expected distance is four times more likely to be selected. Conversely: you would need to have four times larger floorspace in the furthest facility to have equal change of being picked.

The weights are later standardised to add up to 1, so I believe any scale standardisation here would lead to the same result (as the same factor would be applied to each facility). 

However, besides the standardisation, I trust you are also proposing a different functional form, with linearly decaying distance weights. I have tried to avoid this because the slope of the distance decay would be different in each zone, depending on the maximum found distance.  Asymptotic functions like the inverse quardatic or logistic distance functions can give us some more consistency, and I have chosen gravity-looking model as a functional form transport modelers would be familar with.

But the apples and oranges comment is right - we need to argue about the importance of the two different weights and their rate of decay. One could argue that an agent may be indifferent when choosing between very far or between very near locations (the logic behind the logistic formulation).  Or that the distance decay power should be 1 (so that area and distance are simply inversely proportional). Or that a logit choice model would be better.

I am not sure I have a good answer - maybe something to take to domain experts?

pam/variables.py
Theodore-Chatziioannou: updated

pam/variables.py
Theodore-Chatziioannou: added argument overrides on class initialisation (subsequently stored as internal class properties)

pam/core.py
```diff
@@ -306,7 +307,10 @@ def sample_locs(self, sampler):
                             area=act.location.area,
                             loc=sampler.sample(act.location.area, target_act)
                         )
-                        unique_locations[(act.location.area, target_act)] = location
+                        if target_act in variables.LONG_TERM_ACTIVITIES:
```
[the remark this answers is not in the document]
Theodore-Chatziioannou: enabled long_term_activities specification as argument to `sample_locs`

pam/core.py
```diff
@@ -317,6 +321,66 @@ def sample_locs(self, sampler):
...
+
+                    elif isinstance(plan, activity.Activity):
+                        act = plan
+
+                        # remove "escort_" from activity types.
+                        # TODO: model joint trips
+                        if act.act[:7] == "escort_":
+                            target_act = act.act[7:]
+                        else:
+                            target_act = act.act
```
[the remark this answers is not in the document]
Theodore-Chatziioannou: updated

PR #72, opened 2021-02-19
Support matsimv 12

fredshone:
FYI

Adds support for read/write for matsim xml v12 (just add `version = 12`) to the existing matsim read/write methods. Note that default remains 11 so no breaks...

There is no requirement to bump existing pops up to 12 right now (Andrew).

I've also culled the requirements file. Infact the underlying dependancy tree is the same - so no real change i think - just looks nicer. Kasia has suggested removing the jupyter and ipython reqs to save space but because notebooks are the only real use case at the moment i leave in.

Changes are tested pretty well and i added some better doc strings to the write methods.

I made a read/write consistency test for xml v12 so I'm pretty confident it's working, but I've not tested on matsim yet.

A lot of the extra code is to add `__contains__` (eg is `personA in HouseholdA`) and __eq__ (eg is `personA == personB`) to the core module, which then supports the consistency test and i expect to be useful in future. The equalities are agnostic to person and household IDs - they should decend right down into attributes, then child attributes and so on down to plan components. This means that to get a True equality for two big populations will take a long time (i have not tested) but for most use cases where the populations are small or the inequality closer to the parent then it is quick.


fredshone:
This is updated to also support v11 and v12 matsim route reading.

For now I have opted not top support writting routes back to matsim (though i can see a use case in future).

tests/test_1_core_operations.py
```diff
@@ -233,3 +233,214 @@ def test_population_combine_person_duplicate_key():
...
+    p2 = Person('person_X', attributes={1:1})
+    assert p2 in population
+
+    # diff person
+    p2 = Person('person_A', attributes={1:2})
+    assert p2 not in population
+
+
+def test_population_contains_wrong_type():
+    population = Population()
```
mfitz, line 303: Is the exception expected because the population is empty, or because we're not passing in a `Person` type?

tests/test_1_core_operations.py
```diff
@@ -233,3 +233,214 @@ def test_population_combine_person_duplicate_key():
...
+    with pytest.raises(UserWarning):
+        assert None in population
+
+
+def test_hh_contains_person():
+    hh = Household('A', attributes={1:1})
+    p = Person('person_A', attributes={1:1})
+    hh.add(p)
+
+    # same person
```
mfitz, line 313: Again, I'd go with separate tests here

tests/test_3_parse_matsim.py
```diff
@@ -84,3 +84,46 @@ def test_remove_pt_interactions():
...
+    assert legs[1].service_id == 'city_line'
+    assert legs[1].route_id == 'work_bound'
+    assert legs[1].o_stop == 'home_stop_out'
+    assert legs[1].d_stop == 'work_stop_in'
+    assert legs[1].network_route == None
+
+
+def test_fail_v12_plus_attributes():
+    with pytest.raises(UserWarning):
+        population = read_matsim(test_tripsv12_path, attributes_path='fake', version=12)
```
mfitz, line 124: Could we make a stronger assertion about what has happened? Maybe assert on the expected message in the exception, or possibly use a custom exception?

tests/test_8_write.py
```diff
@@ -19,14 +19,15 @@
 
 def test_write_plans_xml(tmp_path, population_heh):
```
mfitz, line 20: I think there are a couple of tests missing here:

1 - Test XML file created is valid against the DTD (grab a local copy of the [DTD](https://www.matsim.org/files/dtd/population_v6.dtd), see [here](https://github.com/arup-group/genet/blob/master/tests/test_outputs_handler_matsim_xml_writer.py#L44) for an example of how to perform the validation)

2 - Test XML file has the expected content, based on the population we've used to create it

tests/test_8_write.py
```diff
@@ -19,14 +19,15 @@
...
 
 
 def test_write_plans_gzip(tmp_path, population_heh):
     location = str(tmp_path / "test.xml.gz")
-    write_matsim_plans(population_heh, location=location, comment="test")
-
+    write_matsim_plans(population_heh, path=location, comment="test")
     expected_file = "{}/test.xml.gz".format(tmp_path)
     assert os.path.exists(expected_file)
     # TODO make assertions about the content of the created file
```
mfitz, line 33: Yes, that would be nice :-P

tests/test_8_write.py
```diff
@@ -52,7 +53,7 @@ def test_write_attributes_gzip(tmp_path, population_heh):
 
 def test_write_read_continuity_xml(tmp_path, population_heh):
     plans_location = str(tmp_path / "test_plans.xml")
-    write_matsim_plans(population_heh, location=plans_location, comment="test")
+    write_matsim_plans(population_heh, path=plans_location, comment="test")
```
mfitz, line 56: Could you use your new population equality function in the assertion for the round trip serialisation/deserialisation here?

tests/test_8_write.py
```diff
@@ -92,9 +93,52 @@ def test_read_write_read_continuity_complex_xml(tmp_path):
     assert complex_plan_in == complex_plan_out
 
 
+def test_write_plans_xml_v12(tmp_path, population_heh):
+    location = str(tmp_path / "test.xml")
+    write_matsim_v12(population=population_heh, path=location, comment="test")
+    expected_file = "{}/test.xml".format(tmp_path)
+    assert os.path.exists(expected_file)
+    # TODO make assertions about the content of the created file
```
mfitz, line 101: Same comment as before - tests for XML validity and then content would be handy

tests/test_8_write.py
```diff
@@ -92,9 +93,52 @@ def test_read_write_read_continuity_complex_xml(tmp_path):
...
+
+
+def test_write_plans_xml_v12_gzip(tmp_path, population_heh):
+    location = str(tmp_path / "test.xml.gz")
+    write_matsim_v12(population=population_heh, path=location, comment="test")
+    expected_file = "{}/test.xml.gz".format(tmp_path)
+    assert os.path.exists(expected_file)
+    # TODO make assertions about the content of the created file
```
mfitz, line 109: Here we should maybe check it's a valid gzip file?

mfitz (changes requested):

pam/write.py
```diff
@@ -164,7 +256,7 @@ def write_matsim_plans(population, location, comment=None):
 
     for hid, household in population:
         for pid, person in household:
-            person.attributes["hid"] = hid  # force add hid as an attribute
+            # person.attributes["hid"] = hid  # force add hid as an attribute
```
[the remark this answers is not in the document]
fredshone: Thanks for the reminder. I've now made adding the household key to person attributes a default behaviour for both v11 and v12. 

This is maybe useful in future so that we can recover hh attributes from matsim populations (we currently don't use a household representation in matsim).

tests/test_1_core_operations.py
```diff
@@ -233,3 +233,214 @@ def test_population_combine_person_duplicate_key():
...
+    hh2 = Household('A', attributes={1:1})
+    p2 = Person('person_A', attributes={1:1})
+    hh2.add(p2)
+    assert hh2 in population
+
+    # same hh and person diff index for hh
+    hh2 = Household('X', attributes={1:1})
+    p2 = Person('person_A', attributes={1:1})
+    hh2.add(p2)
+    assert hh2 in population
```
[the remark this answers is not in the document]
fredshone: yes I know :( I have split them up.

tests/test_1_core_operations.py
fredshone: ...because not passing person type. This seems broadly in line with python although not very safe.

tests/test_1_core_operations.py
fredshone: yes

tests/test_3_parse_matsim.py
fredshone: I agree - but i think this will be premature optimisation at present.

tests/test_8_write.py
fredshone: added

tests/test_8_write.py
fredshone: you are right but maybe next time

tests/test_8_write.py
fredshone: i think this is covered ok by read write continuity tests

tests/test_8_write.py
fredshone: done

tests/test_1_core_operations.py
fredshone: done

mfitz (approved):

Issue #67, opened 2021-02-15
save a facility sampler


akay-arup:
FSs take a while to build, so may be good to be able to create a way to save a pre-baked FS and then importing it into a synthesis pipeline. Currently `FacilitySampler ` class is a generator, which can't be pickled. Consider adding a FacilitySampler.dump() and .load() methods?

Theodore-Chatziioannou:
added export/load functionality

akay-arup:
Missed the boat on closing, but have been testing this today. So far, it seems to save about a minute compared to building from scratch but a large % change. Build from scratch ~200+ seconds, build from pickle ~110 seconds.

PR #73, opened 2021-02-23
enable exporting / loading of joined facility-zone geodataframe

Theodore-Chatziioannou:
Responds to this request: https://github.com/arup-group/pam/issues/67

Enables exporting/loading the spatially joined facilities-zones geodataframe ("activity_areas") from/to the FacilitySampler. By using this method, we can then save time across sampling iteration of the same data, by avoiding the expensive joining part.

To use it, build the facility sampler as usual, and the export the activity areas with `facility_sampler.export_activity_areas(filepath = "activity_areas.pkl")`. Then, next time we initiate/build the facility sampler we can pass the `activity_areas_path` argument to directly read the activity areas into the object (ie `facility_sampler = `FacilitySampler(....., activity_areas_path="activity_areas.pkl").

@akay-arup  Please let me know how much of a time improvement this PR gives.

### DIFF
--- a/pam/samplers/facility.py
+++ b/pam/samplers/facility.py
@@ -11,6 +11,7 @@ from pam import variables
 
 import pandas as pd
 import numpy as np
+import pickle
 from types import GeneratorType
 
 class FacilitySampler:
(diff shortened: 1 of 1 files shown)


akay-arup:
I will run a timed test to execute the sampler for TII and then compare to load + rebuild and get back to you

fredshone (approved):
Great no complaints. To simplify this process in future i am thinking to add the sjoin and tranformers to OSMOX - such that the facility sampler format (which is currently tabular) will include fields for area allocation in addition to all the facility attributes.

The workflow will then be OSMOX CLI -> pam.samplers.facility.FacilitySampler()

PR #79, opened 2021-03-17
Add notebook smoke tests

fredshone:
notebook smoke test passing on my machine.

One hack is that `notebooks/PAM_stability.ipynb` has 2 pip installs within the notebook (seaborn and scipy). I am reluctant to make them a project requirement for one notebook only.

sorry @Theodore-Chatziioannou  I added you by mistake! No need to check this


mfitz (approved):
Lovely stuff! Now that all the notebook errors are fixed, I've added the notebook smoke test to the CI build. The build now takes more than 5 minutes longer, but I think it's worth it:

Screen Shot 2021-03-18 at 01 43 03

PR #81, opened 2021-03-18
Remove unnecessary whitespace from notebook smoketest summary

mfitz:
Extremely trivial and picky, but the extra whitespace was bothering me. Also did some refactoring to make `main` cleaner.

**Before**
Screen Shot 2021-03-18 at 11 48 27

**After**
Screen Shot 2021-03-18 at 13 26 55


fredshone (approved):
beatutiful. Mad thought is that this should be it's own project if we are going to use it in a lot of places.

mfitz:
> beatutiful. Mad thought is that this should be it's own project if we are going to use it in a lot of places.

Yes, I have started thinking about how best to do that.

PR #86, opened 2021-03-29
Example-Create-Population-Londinium

syhwawa:
This notebook demonstrates an example for creating a sample population automatically for the West London area. Any comments or feedback are appreciated.


fredshone (approved):
I've made some additions at the end (please take a look) - but great work - I like the nested loop idea.

please merge into master. I think we will revisit again in future, for example to arrange all these notebooks in a way that is easier for new comers to understand what they are all for. But that can be a future problem.

syhwawa:
> I've made some additions at the end (please take a look) - but great work - I like the nested loop idea.
> 
> please merge into master. I think we will revisit again in future, for example to arrange all these notebooks in a way that is easier for new comers to understand what they are all for. But that can be a future problem.

Thank you for the additions and comments! 

 For sure. I am happy to rearrange it in the future.

PR #88, opened 2021-06-17
fixes multiploygon random point sample

fredshone:
Iseul I add you to this PR for old times sake :)

This deals with the "archepelago" problem - where the existing random point sampler was failing to sample valid points on very sparsely distributed islands.

Modifies the random point sampler to deal with most shapes that can be thrown at it, including points, lines and the multi-varieties. This should now handle islands much better - provided they are multipolygons. Additionally will default to a random point within the shape bounds.

Tested pretty well. Passes coverage and notebook smoke tests.

### DIFF
--- a/pam/samplers/spatial.py
+++ b/pam/samplers/spatial.py
@@ -45,27 +45,70 @@ class RandomPointSampler:
             self.logger.warning(f'Cannot find idx:{idx}, returning None')
             return None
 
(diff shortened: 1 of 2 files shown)


Theodore-Chatziioannou:
thanks, looks great - I like the inclusion of area weights.

Theodore-Chatziioannou (approved):

PR #91, opened 2021-07-19
Pin pandas version to fix unit tests

mfitz:
There has been a change in Pandas version `1.3.0` around the test utility function `assert_frame_equal` (see the [release notes](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.3.0.html#other)). Although I didn't look into exactly why this bug fix breaks some of our tests, regressing to Pandas `1.2.5`, which is the latest version that is older than `1.3`, fixes our tests.

I was able to reproduce the test failures locally via `pip install pandas==1.3.0`, and then fix via `pip install pandas==1.2.5`.

It may be worth someone with knowledge of the code and tests having a quick look to make sure the tests in question are not relying on broken behaviour in `assert_frame_equal()`.

The [build now completes](https://github.com/arup-group/pam/runs/3104892398?check_suite_focus=true).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ geopandas==0.7.0
 ipykernel==5.2.1
 lxml==4.6.2
 matplotlib==3.2.1
-pandas==1.2.5
+pandas==1.3.0
 Rtree==0.9.4
 jupyter==1.0.0
 jupyter-contrib-nbextensions==0.5.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ geopandas==0.7.0
 ipykernel==5.2.1
 lxml==4.6.2
 matplotlib==3.2.1
-pandas>=1
+pandas==1.2.5
 Rtree==0.9.4
 jupyter==1.0.0
 jupyter-contrib-nbextensions==0.5.1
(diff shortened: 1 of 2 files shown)


Theodore-Chatziioannou (approved):

Theodore-Chatziioannou:
@mfitz thanks for this. The tests seem to be breaking due to different column sorting behaviour between pandas 1.3.0 and 1.2.5. The failing tests were comparing identical dataframes, but with different column ordering.

We could either pin pandas to 1.2.5 or align the dataframe columns in the failing tests before checking for equality. I tested the latter option and seemed to pass the tests (it was a small change). Please let me know what you think.

fredshone:
I'd quite like to remove assert_frame_equal()

mfitz:
> @mfitz thanks for this. The tests seem to be breaking due to different column sorting behaviour between pandas 1.3.0 and 1.2.5. The failing tests were comparing identical dataframes, but with different column ordering.
> 
> We could either pin pandas to 1.2.5 or align the dataframe columns in the failing tests before checking for equality. I tested the latter option and seemed to pass the tests (it was a small change). Please let me know what you think.

I prefer fixing the tests and then pinning to the latest pandas @Theodore-Chatziioannou

mfitz (commented):
Awesome!

fredshone (approved):

PR #106, opened 2021-09-21
Consolidate write_travel_diary

JosePazNoguera:
The write_travel_diary method is basically a copy of the to_csv method.
This commit removes the redundant code. I have compared test_8_write.dy results and they look exactly the same.

### DIFF
--- a/pam/write.py
+++ b/pam/write.py
@@ -432,6 +432,8 @@ def to_csv(
                         'tst': component.start_time,
                         'tet': component.end_time,
                         'duration': str(component.duration),
+                        'hzone': hh.location.area,
+
                     }
                     # if component.start_location.area is not None:
                     #     leg_data['start_area'] = component.start_location.area
--- a/pam/write.py
+++ b/pam/write.py
@@ -30,35 +30,36 @@ def write_travel_diary(
 	:param plans_path: str path
 	:param attributes_path: str path
 	"""
-    record = []
-    for hid, pid, person in population.people():
-        for seq, leg in enumerate(person.legs):
-            record.append(
-                {
-                    'pid': pid,
-                    'hid': hid,
-                    'hzone': person.home,
-                    'ozone': leg.start_location.area,
-                    'dzone': leg.end_location.area,
-                    'seq': seq,
-                    'purp': leg.purp,
-                    'mode': leg.mode,
-                    'tst': leg.start_time.time(),
-                    'tet': leg.end_time.time(),
-                    'freq': person.freq,
-                }
-            )
-    pd.DataFrame(record).to_csv(plans_path)
+    to_csv(population,plans_path)
+#    record = []
+#    for hid, pid, person in population.people():
+#        for seq, leg in enumerate(person.legs):
+#            record.append(
+#                {
+#                    'pid': pid,
+#                    'hid': hid,
+#                    'hzone': person.home,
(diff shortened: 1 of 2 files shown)


mfitz (approved):
I love it - delete is the best refactoring!

Can we delete the commented out code though?

JosePazNoguera:
Commit ae8d1a7 builds on the work @mfitz did. Apart from renaming the two tests, it looks at the legs.csv file not the activities.csv file. 
The first test checks whether the file exists. The second test compares the content of legs.csv (created by the to_csv method) against the output created by the master branch (using the previous user-defined method). The test fails because the order of the columns has changed, there are a couple extra columns and the formatting of the time is different when you use the to_csv method.

pam/write.py
```diff
@@ -30,35 +30,9 @@ def write_travel_diary(
...
-                    'mode': leg.mode,
-                    'tst': leg.start_time.time(),
-                    'tet': leg.end_time.time(),
-                    'freq': person.freq,
-                }
-            )
-    pd.DataFrame(record).to_csv(plans_path)
-
+    to_csv(population,plans_path)
     if attributes_path:
```
fredshone, line 34: remove attributes path option altogther, the `to_csv` function will output attributes in any case

pam/write.py
```diff
@@ -458,6 +432,8 @@ def to_csv(
                         'tst': component.start_time,
                         'tet': component.end_time,
                         'duration': str(component.duration),
+                        'hzone': hh.location.area,
```
fredshone, line 435: good

tests/test_8_write.py
```diff
@@ -257,13 +257,52 @@ def test_read_write_v12_consistent(tmp_path):
...
+        'people.csv',
+        'people.geojson',
+    }
+    assert set(os.listdir(location)) == expected_files, \
+        "Directory does not contain expected file list '{}'".format(expected_files)
+
+
+def test_write_travel_diary_creates_expected_legs_csv_file(tmp_path, population_heh):
+    location = os.path.join(tmp_path, "test_plans")
+<<<<<<< HEAD
```
fredshone, line 284: remove these `<<<<<<<<<`

fredshone: etc

fredshone (changes requested):
there are some lingering `>>>>>>>>>` etc which are causing tests to fail

JosePazNoguera:
Apologies for the V2 and V3 commits. I didn't realize a small typo and then created a merge conflict .... 
Just ignore them. The relevant commit is the latest one "Fixed Typo"

pam/write.py
JosePazNoguera: OK

fredshone:
Great effort - I've made some changes that I'd like you do consider

1. removed `write_travel_diary` entirely - I noticed it was not in use in any of the examples - if someone complains we can revert but i think this will be fine as the required funcionality is in `to_csv`.
2. removed the hh.location.area field from the legs outputs - this information was duplicated in the hh output
3. fixed tests to suit
4. i refactored the `write_population_csv` to use the `to_csv` logic. Any changes for future improvements are all in one place now. I also renamed to `write_population_csvs`. I also confirmed this was not in use in the examples and I fixed (deleted) the required tests for a slight loss in coverage that we can address in future
5. Some minor formatting

fredshone (approved):

PR #107, opened 2021-09-28
Optimise read

Theodore-Chatziioannou:
Reading travel diary data into PAM is a frequently-used procedure, which however can be a speed bottleneck when working with large populations. This PR is a code refactoring which achieves a 70%-90% reduction in pam.read methods' run times:

read_comparison

The process for identifying the bottlenecks and speeding up the code is documented (with some examples) separately [here](https://github.com/arup-group/speed-profiler-cml/blob/main/pam/README.md).

Next step: speed up facililty sampling


mfitz (approved):
Really good stuff @Theodore-Chatziioannou!

As far as a more systematic way to analyse performance, there are a couple of connected areas here:

#### Benchmarking performance
By this I mean measuring and validating the overall performance of some part of the API, e.g. loading travel diary data. There is a fairly crude shell script `scripts/performance/population-builder-profile.sh` and a benchmark file at `scripts/performance/simple-pam-data-x10000-benchmark.dat` that does this kind of thing for creating a population, using `mprof` to measure the running time and memory usage. It's not an attempt to understand where the code spends its time, just a basic mechanism for measuring performance and optionally comparing to a known benchmark. I think it's a bit neglected, but I just ran it and found the performance is much faster now than the benchmark, although memory usage has doubled:

Screen Shot 2021-09-28 at 17 30 10

You could do something similar for any part of the API you feel like isolating in a small app (`scripts/performance/activity_loader.py` is the one I wrote for population building) that you can then profile. With Puma we've included something like that [in the GitHub Actions CI build](https://github.com/arup-group/puma/blob/master/.github/workflows/build_pipeline.yml#L35); the intention is that the build will break if memory usage increases, so it was done as much as a safeguard as anything.
 
#### Performance profiling
This is the kind of thing you've been doing here with the line profiler - analysing where the code spends its time. I don't know if that's a thing we need to do systematically; it seems more like something we do as and when needed when we know there is a performance bottleneck that we'd like to address. You could wrap the line profiling in a little script that captures the outputs, maybe, but it isn't the kind of thing I would see us doing in the CI build, for example.

Did you find any friction points in your profiling methodology while you worked on this? It certainly seems to have helped you to improve the performance of this bit of the API. Or is it that you want to do the same sort of thing for other parts of/all of the API?

pam/read.py
```diff
@@ -453,20 +465,44 @@ def add_persons_from_trips(
...
-        
+        if pid in household.people:
+            continue
+        person = core.Person(
+            pid,
+            home_area=hh_person_data.iloc[0].to_dict().get("hzone"),
+            )
+        household.add(person)
+
+def reindex_df_dict(
```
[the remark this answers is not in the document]
akay-arup: This is great and powerful. I would prefer a slightly more descriptive name like `hh_df_to_dict` -- especially because there isn't an explicit reindex inside the function. Likewise, perhaps `group`  could be renamed `pid_data` or `person_activities` or similar?

pam/read.py
```diff
@@ -453,20 +465,44 @@ def add_persons_from_trips(
         return None
 
     logger.info("Adding persons from trips")
-    for hid, hh_data in trips.groupby('hid'):
+    # for hid, hh_data in trips.groupby('hid'):
```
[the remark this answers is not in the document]
Theodore-Chatziioannou: commented lines removed

pam/read.py
```diff
@@ -495,17 +531,19 @@ def tour_based_travel_diary_read(
...
+
+    if sort_by_seq:
+        trips = trips.sort_values(['hid','pid','seq'])
+
+    trips_dict = reindex_df_dict(trips, 'hid', 'pid') # convert to dict for faster indexing
     
     for hid, household in population:
         for pid, person in household:
-            person_trips = trips.loc[(trips.hid == hid) & (trips.pid == pid)]
+            person_trips = trips_dict.get(hid, {}).get(pid, {})
```
fredshone, line 536: note that this returns a dataframe or a dict if not found - not great to have two types, perhaps switch `reindex_df_dict` to use `values_dict=True` and refactor remaining?

pam/read.py
```diff
@@ -453,20 +465,44 @@ def add_persons_from_trips(
...
+
+    The dictionary structure allows for much faster access to a person's data.
+    :params pd.DataFrame df: the pandas dataframe to reindex
+    :params str key_hh: the household key column name
+    :params str key_person: the person key column name
+    :params boolean values_dict: whether to convert the person data to a dictionary as well
+    """
+    df_dict = {x:{} for x in df[key_hh].unique()}
+    for (hid, pid), group in df.groupby([key_hh,key_person]):
+        if values_dict:
```
fredshone: where you thinking to also use this? if not, maybe remove.

I think it makes good sense to switch to dicts - dictionaries of dictionaries of dataframes is a bit yucky

pam/read.py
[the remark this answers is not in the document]
Theodore-Chatziioannou: renamed

pam/read.py
Theodore-Chatziioannou: yep, I have already removed the `values_dict` step. The intention is, as you say to have the same data type throughout, but it was slow and used an alternative.

pam/read.py
Theodore-Chatziioannou: I think the empty dictionary would then only be used to call the `person.stay_at_home()` method, then move to the next person. But agree, it's not good practice - perhaps I could replace with `person_trips = trips_dict.get(hid, {}).get(pid, pd.DataFrame())`?

Theodore-Chatziioannou:
thanks @mfitz, your comment is very helpful. There were no particular friction points, but I was thinking that it need to be complemented by something more high-level, which can be run routinely. I really liked the pam memory benchmarking script (I wasn't aware of it), and it seems to be doing exactly that. I could work towards adding more modules to it.

PR #109, opened 2021-09-30
Update population building performance profiling script

mfitz:
@Theodore-Chatziioannou's recent profiling work reminded me that the performance measuring scripts and benchmarks for population building needed a bit of love:

- Benchmark was out of date
- Script was difficult to use and relied on using your own local population data
- `mprof` was not included in the project dependencies

After addressing these issues and running the new `scripts/performance/population-builder-performance-benchmark.sh` script, the generated report looked like:

Screen Shot 2021-09-30 at 23 50 43

Screen Shot 2021-09-30 at 23 51 01

I have taken the decision not to add any instructions about this to the `README.md`, on the basis that using this script is quite a niche thing to want to do. I'm happy to receive alternative opinions; perhaps some documentation on this would have been useful to you during your recent performance tuning work, @Theodore-Chatziioannou? Should we include some documentation as a pointer to anyone that may want to do something similar for a different part of the PAM library? Or should we extend our performance testing to include more parts of the library, rather than just population building, and maybe even add it to the CI build, as we have done with [Puma](https://github.com/arup-group/puma/blob/master/.github/workflows/build_pipeline.yml#L35)?


Theodore-Chatziioannou:
thanks @mfitz . Agree that documentation does not need to be public for this. Perhaps you could simply copy your comments on the two PRs (ie how it works and the different kinds of profiling) to [this](https://github.com/arup-group/speed-profiler-cml/tree/main/pam) separate repo, as your work is helpful reference and could be reused?

Theodore-Chatziioannou:
Worked fine for me and was straightforward to run. Windows users need to run it via the Windows Linux Subsystem. 

Do you think profiling should look into inputs of different size? I tried this idea [here](https://github.com/arup-group/speed-profiler-cml/blob/5af023a399c98ee8c0ec055148a2f28c90796e36/pam/01_read.py#L19-L34), where we randomly sample n agents and rerun the process (with n increasing every time). However, it turned out that runtime scales linearly (more or less) with the size for the read functions, so it was not adding much value.

I would like to include facility sampling to the profiling - it is the second major bottleneck after population loading.

Theodore-Chatziioannou (approved):

fredshone (approved):
LGTM, I see adequate docs here:
````
print("Performance report written to {}".format(os.path.abspath('pam-performance-report.html')))```

mfitz:
> thanks @mfitz . Agree that documentation does not need to be public for this. Perhaps you could simply copy your comments on the two PRs (ie how it works and the different kinds of profiling) to [this](https://github.com/arup-group/speed-profiler-cml/tree/main/pam) separate repo, as your work is helpful reference and could be reused?

I'll add something to that repo @Theodore-Chatziioannou 👍

mfitz:
> Worked fine for me and was straightforward to run. Windows users need to run it via the Windows Linux Subsystem.

Awesome!

> 
> Do you think profiling should look into inputs of different size? I tried this idea [here](https://github.com/arup-group/speed-profiler-cml/blob/5af023a399c98ee8c0ec055148a2f28c90796e36/pam/01_read.py#L19-L34), where we randomly sample n agents and rerun the process (with n increasing every time). However, it turned out that runtime scales linearly (more or less) with the size for the read functions, so it was not adding much value.

Different sized inputs are definitely helpful when you are profiling a specific thing. There is a balance here between providing the developer with everything you think they might potentially need, versus giving them enough flexibility to use the tools in whatever way makes sense for their use case, and I think it can be quite tough to pre-emptively provide all the data anyone will ever need for their future, unknown use case. I lean towards not doing this for the population builder stuff, but if it's useful for the work you have just done and plan to do for other parts of the API, I am open to discussion.

In the population builder case, you can pass whatever population data you want to the perf testing script, so if you have a useful dataset locally you can use it, although you'd probably need to dig around in the scripts to figure out how. That's why I added some default test population data, but I don't see it as a comprehensive performance testing dataset or anything.

On a semi-related note - benchmarks. With Puma, we found that the application profiled differently depending on the host OS, so you need multiple "default" benchmarks too, e.g. an OS X benchmark, an Ubuntu benchmark, and so on. I'm not especially bothered about that here; I imagine a situation where you are about to start some performance tuning work on a particular corner of the API. The first thing to do would be to run the perf script a few times with no benchmark, and select one of those runs to use as your benchmark. You would then start your perf tuning work, using the script and benchmark as a feedback mechanism to guide you while you make changes. Hence, it's not a pre-baked benchmark or set of benchmarks that is useful so much as the ability to create and then use your own benchmarks. Again though, this is not at all obvious, so I'm talking myself into writing some documentation...

If we wanted to add perf testing to the CI build, we'd obviously need pre-baked benchmarks to catch any performance regression at build time.

> 
> I would like to include facility sampling to the profiling - it is the second major bottleneck after population loading.

We can do that. I guess we're approaching a decision point about whether or not to create another ad-hoc app and script, or to consolidate into a single, generic thing. I can see pros and cons for each. With a bunch of separate apps/scripts, you know exactly which area of the API is running slowly if something starts to perform badly against previous benchmarks, but then you can end up with a bunch of duplication and bloat. Facility sampling is a good next use case for digging in to that decision - happy to discuss outside of this PR. 

At the very least, we can make the `scripts/performance/population-builder-profile.sh` shell script that calls `mprof` and makes/compares to benchmarks generic enough that the app to be profiled (and its command line params) is a parameter, rather than the current hardcoding to `activity_loader.py`. That seems like a quick, easy and useful change. The typical workflow would then be:

- write a small python app to exercise the bit of the API you want to performance tune
- generate a benchmark/some benchmarks for that app using a generic profiling script
- start your perf tuning changes, using the script and benchmarks in a feedback loop while you work

Would that work for you initially @Theodore-Chatziioannou? We can make it smoother as we go.

PR #110, opened 2021-10-04
Add zone for londinium population

syhwawa:
We have frequently used a zoning system when we created a population for ABM projects. To make the Loninium population more realistic, I added zones attributes including MOSA_name and LAD for each agent base on their first activity location.


fredshone (approved):
...but also please consider removing all the POINT print statements

Theodore-Chatziioannou:
`lsoas_clipped` and `zones` are essentially the same file, but are used interchangeably in the notebook - suggest you only keep one of the two

Theodore-Chatziioannou:
in this cell:
```
for i in range(0, 306):
    print(facility_sampler.sample(i, 'work'))
```
, I guess the purpose is to demonstrate that the facility sampler is working? If so, just printing the first 5 points would suffice, no need to repeat the process so many times

PR #113, opened 2021-11-19
facility sampler speed improvements

Theodore-Chatziioannou:
Small refactoring with large impact on the facility sampler build speed. Testing with a fairly large facilities/zones dataset, it shaved off more than 90% of the `FacilitySampler.build_facilities_sampler` method runtime.

The spatial join between zones and activities is still slow, but I don't think it can be improved further. However, the already implemented `export_activity_areas` and `load_activity_areas` methods allow to skip that step in subsequent iterations. Another option could also to make the spatial join part of OSMOX (as in most cases it only needs to happen once per project).


fredshone (approved):
nice

PR #118, opened 2022-01-04
Reproducible sampling

YannisZa:
I have modified the sampling scripts `attributes.py`, `basic.py`, `facility.py`, and `spatial.py` to pass a seed parameter to fix the pseudorandom number generator's seed. The default is None which does not fix the seed. I did this so that people can generate reproducible sampled facility locations in `pam`.


YannisZa:
Just realised that the commits I pushed in another branch are also pushed in this branch.

fredshone:
> Just realised that the commits I pushed in another branch are also pushed in this branch.

do you want to proceed?

YannisZa:
> 

Shall I close and create a new PR?

Issue #117, opened 2022-01-04
Make sample_locs in Population class reproducible 


YannisZa:
Fix random seed when sampling locations in [def sample_locs](https://github.com/arup-group/pam/blob/fc59b1f386bb000e38b1ada8af5f26071262c8ea/pam/core.py) so that process can be reproduced.

YannisZa:
There is a caveat to the above. If you want to sample facilities you need to run the following together
```
facility_sampler = facility.FacilitySampler(
    facilities=cambridge_facilities.assign(weight1=1),
    zones=wards,
    build_xml=True,
    fail=False,
    random_default=False,
    seed=seed_num
)

facility_sampler.clear()

# Sample facility locations for each activity
population.sample_locs(facility_sampler)
```

If you run ```population.sample_locs(facility_sampler)``` on its own reproducibility will be lost. This is because the facility sampler uses an infinite yield generator to shuffle facilities as shown in functions `build_facilities_sampler` and `inf_yielder_simple` of [this script](https://github.com/arup-group/pam/blob/main/pam/samplers/facility.py). Therefore, every time you call `facility_sampler.sample(zone,activity)` the `next` item of the infinite generator will be returned. 

It would be good to investigate how to ensure reproducibility inside an infinite generator - although this is not a pressing issue.

fredshone:
Understood and excellent point, maybe something as simple as renaming/wrapping `.clear()` as `.reset()` would help understanding and correct use. Next time we add an example notebook we can also make a seeded example.

A stronger step would be to have the `.sample_locs(Sampler)` methods call reset on the sampler prior to use.

More generally the reason we use an object for the sampler (and therefore have persistence and potential seeding mistakes) is because it takes AGES to build the sampler object, so we were keen to reuse them once in memory. But we plan to move the build (essentially just the assignment of facility locations to zones) into the prior step (OSMOX) - at which point we could drop the Sampler object altogether, ie:

```
sampler = Sampler.build(data)
population.sample_locs(sampler)
```
--->
```
population.sample_locs(data)
```

But i am just thinking out loud - no actions required immediately

PR #119, opened 2022-01-04
Reproducible sampling

YannisZa:
Passed seed parameter in `attributes.py`, `basic.py`, `facility.py`, and `spatial.py` samplers to ensure that sampling facility locations is reproducible. Default seed is `None` which does not fix the pseudonumber generator's seed.

### DIFF
--- a/pam/samplers/attributes.py
+++ b/pam/samplers/attributes.py
@@ -1,6 +1,4 @@
-# from random import random
 import random
-
 
 def bin_integer_transformer(features, target, bins, default=None):
     """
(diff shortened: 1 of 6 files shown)


pam/samplers/attributes.py
```diff
@@ -1,4 +1,5 @@
-from random import random
+# from random import random
```
fredshone, line 1: delete me

fredshone (approved):
lovely PR - really appreciate the tests

YannisZa:
This PR addresses issue #117

PR #122, opened 2022-02-01
Allow building GeoDataFrame for people WFH

KasiaKoz:
Discovered through working with SF model https://github.com/eqasim-org/california

If someone stays home all day, their GeoDataFrame cannot be build with the following error message:

Screenshot 2022-02-01 at 09 46 11

Fix is a mostly empty GeoDataFrame with `Point` geometry. That way the agent shows up on a map. I decided against adding things like duration - there is no trip so no duration in my mind.

Screenshot 2022-02-01 at 09 47 28


fredshone (changes requested):
...given that this is called `build_person_travel_dataframe` - would prefer this to just have legs/trips in (and therefore not show non-travellers.

The pattern elsewhere is to treat leg and activity outputs seperately.

However, no complaints if you want to rename this to something like `build_person_plan_df` for example?

PR #124, opened 2022-02-14
fix complex sampler

fredshone:
just a little debug that was required to run the fac sampler. Not sure why

### DIFF
--- a/pam/samplers/facility.py
+++ b/pam/samplers/facility.py
@@ -297,6 +297,8 @@ def inf_yielder_weighted(candidates, weights, transit_distance, max_walk, transi
                     weights * variables.SMALL_VALUE, # if no alternative is found within the acceptable range, the initial weights will be used
                     weights
                 )
+            else:
+                weights = weights.values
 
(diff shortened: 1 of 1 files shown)


Theodore-Chatziioannou:
do you recall what was the error? I had added the `.values` property so that we divide objects of the same type (numpy array).

fredshone:
yes - weights was already an array.

From this block pressumably?

```           
if isinstance(transit_distance, pd.Series) and mode in transit_modes:
    weights = np.where(
        transit_distance > max_walk,
        weights * variables.SMALL_VALUE, # if no alternative is found within the acceptable range, the initial weights will be used
        weights
   )
```

Theodore-Chatziioannou:
ok, so there is a code path that casts weights to arrays and one where they are kept as Series (which I think will crash after your update?). Should we cast to array in any case then?

fredshone:
sure - let me re push

Theodore-Chatziioannou (approved):

PR #129, opened 2022-03-09
adds population sampler

fredshone:
adds population (freq) sampler function for convenience (I use it a lot)

### DIFF
--- a/pam/samplers/population.py
+++ b/pam/samplers/population.py
@@ -1,0 +1,59 @@
+from pam.core import Population
+from copy import deepcopy
+from pam.samplers.basic import freq_sample
+
+
+def sample(
+    population: Population,
+    sample: float,
+    seed: int=None,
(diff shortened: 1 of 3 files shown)


pam/samplers/population.py
```diff
@@ -0,0 +1,59 @@
...
+from copy import deepcopy
+from pam.samplers.basic import freq_sample
+
+
+def sample(
+    population: Population,
+    sample: float,
+    seed: int=None,
+    verbose: bool=False
+    ):
```
Theodore-Chatziioannou, line 11: ) -> Population :

Theodore-Chatziioannou (approved):
nice tests

Theodore-Chatziioannou:
only one comment: could you clarify the hierarchy of frequency fields somewhere? What happens if there are household, person, and leg weights in a population?

PR #134, opened 2022-03-29
Handle WFH Persons when building travel `GeoDataFrame`

KasiaKoz:
Handles an issue discovered through working with SF model https://github.com/eqasim-org/california

If someone stays home all day, their GeoDataFrame cannot be build with the following error message:

Screenshot 2022-02-01 at 09 46 11

Persons not traveling get `None` instead of a `GeoDataFrame` and that is handles later when building a Household and Population level `GeoDataFrame`s, where they are not included.

This is a different, improved behaviour from the previous attempt in https://github.com/arup-group/pam/pull/122


pam/core.py
```diff
@@ -628,13 +629,14 @@ def build_travel_geodataframe(self, **kwargs):
...
-        gdf = gdf.sort_values(['pid', 'seq']).reset_index(drop=True)
-        return gdf
+            if _gdf is not None:
+                _gdf['hid'] = self.hid
+                if gdf is None:
+                    gdf = _gdf
+                else:
+                    gdf = gdf.append(_gdf)
+        if gdf is not None:
+            return gdf.sort_values(['pid', 'seq']).reset_index(drop=True)
```
fredshone, line 639: can `build_travel_geodataframe` return an empty geodataframe rather than None please? Then can loose the nasty logic tree.

fredshone (changes requested):

PR #136, opened 2022-04-05
read non-selected plans

Theodore-Chatziioannou:
Enables PAM to read non-selected plans from MATSim plan outputs. This option is enabled by setting the `keep_non_selected` argument of the `read.read_matsim` as True. The non-selected plans are stored as a list of plans under person.plans_non_selected.

```
population = read.read_matsim(
    plans_path = "output_plans.xml.gz",
    crop = False,
    keep_non_selected = True
)

population.random_person().plans_non_selected
```
>>returns: [<pam.Plan>, <pam.Plan>, etc]

### DIFF
--- a/pam/core.py
+++ b/pam/core.py
@@ -701,6 +701,7 @@ class Person:
         self.person_freq = freq
         self.attributes = attributes
         self.plan = activity.Plan(home_area=home_area)
+        self.plans_non_selected = []
         self.home_area = home_area
         self.vehicle = None
         if vehicle:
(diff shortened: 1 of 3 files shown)


fredshone:
I'd quite like to future proof this (for the future that I imagine) by storing all plans as a list:

`Person.plans = [Plan, Plan...]`

then make `plan` and 'unselected_plans` methods that extract as expected based on scores. We can keep it backward compatible using @property decorator.

Where there are no scores and more than one plan (ie where there is ambiguity then we can have an exception).

My rationale is that this will be easier to understand and if we do rescheduling within this framework it would make a nice api, eg:

```
prev_best = person.plan
person.score(ChatzShoneScorer)  # rescore
if not person.plan == prev_best:
 etc
```
or
```
person.add(PlanMutator(person)).score()
```

fredshone (approved):

gac55 (approved):
Great

PR #139, opened 2022-04-29
Home area fix

Theodore-Chatziioannou:
Tries to bring more consistency when assigning home locations to persons. If a household location exists, it will get assigned as the `home_location` for every household member. This ensures that all household members will have the same home location, and that agents without any activities ("stay at home") will still have a location.

If there is no household location, PAM will infer the home location from the trips/activities table.

### DIFF
--- a/pam/activity.py
+++ b/pam/activity.py
@@ -2,7 +2,9 @@ from datetime import datetime
 from datetime import timedelta
 import logging
 from copy import copy
+from sunau import Au_read
 
+from pam.location import Location
 from pam.plot import plans as plot
 import pam.utils
 import pam.variables
@@ -12,9 +14,16 @@ from pam.variables import END_OF_DAY
 
 class Plan:
 
-    def __init__(self, home_area=None, freq=None):
+    def __init__(self, home_area=None, home_location:Location=None, home_loc=None, freq=None):
         self.day = []
-        self.home_location = Location(area=home_area)
+        if home_location:
+            self.home_location = home_location
+        else:
+            self.home_location = Location()
+        if home_area:
+            self.home_location.area = home_area
+        if home_loc:
+            self.home_location.loc = home_loc
         self.logger = logging.getLogger(__name__)
         self.plan_freq = freq
         self.score = None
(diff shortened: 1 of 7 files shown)


fredshone:
...started with a minor change that cascaded pretty bad.

TLDR:

```
def test_home_location_consistency_between_person_and_plan(trips, persons_attributes, hhs_attributes):
    """
    Note that this works because people share a location object with their plans.
    """
    population = load_travel_diary(
        trips=trips,
        persons_attributes=persons_attributes,
        hhs_attributes=hhs_attributes
        )
    hh = population[0]
    person = hh[0]
    assert person.home.area == "Harrow"
    assert person.plan.home == "Harrow"
    person.home.area = "Test"
    assert person.home.area == "Test"
    assert person.plan.home == "Test"


def test_home_location_consistency_between_hhs_and_persons_when_changing_hh_area(trips, persons_attributes, hhs_attributes):
    population = load_travel_diary(
        trips=trips,
        persons_attributes=persons_attributes,
        hhs_attributes=hhs_attributes
        )
    hh = population[0]
    person = hh[0]

    assert hh.location.area == "Harrow"
    assert person.home == "Harrow"
    hh.location.area = "Test"
    assert person.home == "Harrow"  # no change
    hh.set_area("Test")
    hh.set_loc("Test")
    assert person.home == "Test"
    assert hh.location == person.home
```

Convention is that Persons and their Plan with share a Location object. But Households will not. Not sure if this is really better.

I also found some bugs in the read method.

Smoke tests ok on my machine and I think i've avoided breaks but please check your stuff...

pam/activity.py
```diff
@@ -2,7 +2,9 @@
 from datetime import timedelta
 import logging
 from copy import copy
+from sunau import Au_read
```
Theodore-Chatziioannou, line 5: is this needed?

pam/activity.py
```diff
@@ -12,9 +14,16 @@
 
 class Plan:
 
-    def __init__(self, home_area=None, freq=None):
+    def __init__(self, home_area=None, home_location:Location=None, home_loc=None, freq=None):
```
Theodore-Chatziioannou, line 17: `home_location: Optional[Location]`

Theodore-Chatziioannou (commented):
good idea - added two minor comments

fredshone (approved):

PR #141, opened 2022-05-05
Freight pop

divyasharma-arup:
A script designed for freight population creation through the following objects:

1. CreateDistribution to produce distributions of the number of stops and time of delivery.
2. FrequencySampler to sample distributions
3. InputDemand to determine facility density and distance between zones
4. TourSequence that samples destination zones and creates a sequence of these stops based on distance from origin.
5. ActivityDuration that calculates the distance, journey time, and stop time of activities. 
6. TourPlan creates the activity and leg objects that provide a tour plan

An example notebook in the examples folder is provided (17_tour.ipynb). Right now, this uses the HK population data as a test, but will switch this to a Londinium population set if possible. I was running into issues with the facility sampler using London LSOA data and londinium facilities, still need to work to understand what is driving the errors. Please take a look at the notebook, especially in the Activity Modeller and Visualisations sections. Right now, these two parts of the notebook are a mix of code and PAM libraries. I'm not certain if we think it is necessary for this non-library code to shift to a PAM library as well.

Tests for this script can be found in test_9_samplers_tour.py. 

Next Steps from here:
1. Update example notebook to utilise Londinium Data
2. Further testing and documentation (possibly setup sphinx?)
3. Additional functionality to create freight population based on OD flows.


divyasharma-arup:
@fredshone and @Theodore-Chatziioannou , I have updated the example notebook (17_tour.ipynb) to utilise the Londinium datasets and have removed HK datasets from the repo. I've also changed how the od matrix is created in this example notebook to utilise scipy's spatial library. Tests have been included, although only 63% of the code has a test associated with it. Therefore, I anticipate needing to add a few more tests before the code is finalised.

Please let me know your comments.

pam/samplers/tour.py
```diff
@@ -0,0 +1,309 @@
...
+import random
+import pandas as pd
+
+import pam
+from pam.activity import Activity, Leg
+from pam.variables import END_OF_DAY
+from pam.utils import minutes_to_datetime as mtdt
+
+
+class CreateDistribution:
```
fredshone, line 13: don't like this name (it would be a good function or method name but less so a class name).

for example the code would read:
```
CreateDistribution().build_distribution()
```
which doesn't read well. It is convention to use nouns for classes and "doing statements" for methods/functions, for example, something like `DistributionFactory` -> `DistributionFactory().create_distribution()`

This is NOT 🥲 pedantry - it makes the code easier for other people to read and write.

I also think it woulld make sense to include the `build_distribution()` etc in an init method, then add a new method called `sample()`. Such that this class becomes a generic `BespokeDistribution` class, eg:

```
daily_trips_sampler = BespokeFrequencySampler(bins, pivots)
departure_hour = daily_trips.sample()
```
this better matches the existing pattern within the `pam/sampler` module.

pam/samplers/tour.py
```diff
@@ -0,0 +1,309 @@
...
+        """
+        return random.choices(self.distribution, weights=self.frequency, k=1)[0]
+    
+    def samples(self, n=1):
+        """
+        :param n: number of samples to be returned
+        :return: single object or list of objects sampled from distribution
+        """
+        return random.choices(self.distribution, weights=self.frequency, k=n)
+
```
fredshone, line 97: 2 lines between classes and functions (1 between methods within a class is fine)

pam/samplers/tour.py
```diff
@@ -0,0 +1,309 @@
...
+import pam
+from pam.activity import Activity, Leg
+from pam.variables import END_OF_DAY
+from pam.utils import minutes_to_datetime as mtdt
+
+
+class CreateDistribution:
+    """
+    Returns a distribution and corresponding plot based on input hourly demand values.
+    The resulting distribution can be used to sample time of day of activities or number of repetition of activities.
```
fredshone, line 16: ...doesn't have to be used as an hour sampler, please generalise the doc string.

pam/samplers/tour.py
```diff
@@ -0,0 +1,309 @@
...
+        return random.choices(self.distribution, weights=self.frequency, k=1)[0]
+    
+    def samples(self, n=1):
+        """
+        :param n: number of samples to be returned
+        :return: single object or list of objects sampled from distribution
+        """
+        return random.choices(self.distribution, weights=self.frequency, k=n)
+
+class InputDemand:
```
fredshone, line 98: I like this class but I think it should be called `FacilityDensityDsitribution`.  As per previous comments consier adding a sample method and calling this class `FacilityDensityZoneSampler`.

`zone_id` arg needs a bit of better explanation. Definately add to doc string. It is quite comment to make the zone id the dataframe index in other workflows - in which case i think this would break?

pam/samplers/tour.py
```diff
@@ -0,0 +1,309 @@
...
+        d_density = facility_zone[facility_zone['activity']==d_activity].groupby(zone_id).agg({'activity':'count'}).reset_index()
+        d_density['density'] = d_density['activity']/d_density['activity'].sum()
+
+        return o_density, d_density
+
+class TourSequence:
+    """
+    Object samples destinations based on the destination density and develops a tour sequence.
+    """
+
```
fredshone, line 126: remove extra line/s

pam/samplers/tour.py
fredshone: If you make this a sampler you get to decide how specialised it is. You could either do:
```
od_sampler = FacilityDensityZoneODSampler(o_activity="home", d_activity="work")
o = od_sampler.sample_origin()
d = od_sampler.sampler_destination()
```
or, very similarly:
```
od_sampler = FacilityDensityZoneODSampler(o_activity="home", d_activity="work")
o, d = od_sampler.sample_od()
```
or more generally (but requires duplicatiopn of compute for the spatial join):
```
o_sampler = FacilityDensityZoneSampler(activity="home")
o = od_sampler.sample()
d_sampler = FacilityDensityZoneSampler(activity="work")
d = d_sampler.sample()
```
my favourite would be:
```
od_sampler = FacilityDensityZoneSampler()
o = od_sampler.sample(activity="home")
d = od_sampler.sample(activity="work")
proposed_shop_d = od_sampler.sample(activity="shop")
```

pam/samplers/tour.py
```diff
@@ -0,0 +1,309 @@
...
+        """
+        return distance / speed
+
+    def model_stop_time(self, time, maxi, mini): # maxi=3600, mini=600
+        """
+        time in s
+        """
+        return max([mini, min([time, maxi])]) #  100% of estimated journey time
+
+    def activity_duration(self, o_loc, d_loc, end_tm, speed, maxi, mini):
```
fredshone, line 212: `ActivityDuration().sample()`?

fredshone (changes requested):
As a starter I've added annoying comments about naming things. basically classes should be nouns and methods the thing they do. Such that you can read them as `MyClass().do_something()`.

Also the other samplers use a pattern of `Sampler().sample()`. So suggest swapping out some of the bespoke method names for sample, eg `Sampler().sample()`.

pam/samplers/tour.py
divyasharma-arup: This has been renamed to PivotDistributionSampler.

pam/samplers/tour.py
divyasharma-arup: Unfortunately I missed doing this consistently for each class. I'll fix it on the next push along with any other changes requested.

pam/samplers/tour.py
divyasharma-arup: This object has been created, FacilityDensitySampler, and has a similar behaviour to the favourite option here.

pam/samplers/tour.py
divyasharma-arup: Another formatting issue I may need to address on next push.

pam/samplers/tour.py
```diff
@@ -0,0 +1,398 @@
...
+            d_zone = dzone_sampler.dzone_sample(o_zone=o_zone, df_od=df_od, dist_threshold=dist_threshold)
+            d_facility = facility_sampler.sample(d_zone, d_activity)
+            d_seq.append({
+                'stops': j,
+                'destination_zone': d_zone,
+                'destination_facility': d_facility,
+                'distance': ActivityDuration().model_distance(o_loc, d_facility)
+            })
+    
+        d_seq = pd.DataFrame(d_seq)
```
fredshone, line 221: a niggle only - converting something to a df and then immediately extracting a dict is a bit weird.

pam/samplers/tour.py
```diff
@@ -0,0 +1,398 @@
...
+
+        # sort distance: furthest facility to closest facility to origin facility. The final stop should be closest to origin.
+        d_seq = d_seq.sort_values(by=dist_id)
+        d_loc = d_seq['destination_facility'].to_dict()
+
+        return o_zone, o_loc, d_zone, d_loc
```
fredshone, line 227: should this be called `d_zoneS` and `d_locS`? Also, at the moment, `d_zone` is the last zone from the loop, which is then sorted so it's essentially a random zone from the destinatiun zones.

pam/samplers/tour.py
```diff
@@ -0,0 +1,398 @@
...
+        trip_duration = self.model_journey_time(trip_distance, speed)
+        stop_duration = self.model_stop_time(trip_duration, maxi, mini)
+
+        start_tm = end_tm
+        end_tm = end_tm + int(trip_duration/60)
+
+        return stop_duration, start_tm, end_tm
+
+
+class TourPlan:
```
fredshone, line 187: `TourPlanner`?

the `ozone_sampler`, `d_zone_sampler`, `facility_sampler`, `df_od`, `dist_id` can be added to an innit method, eg:

```
class TourPlanner:
    def __init__(self, ozone_sampler, d_zone_sampler, facility_sampler):
        self.ozone_sampler = ozone_sampler
... 
```

and a niggle: `df_od` ~> `od_matrix` (it's not necessary to say what data type something is in the name - better include in the doc string and type checking which we can implement later).

fredshone: you can then add a catch all method:?

```
def apply(self, agent, stops, dist_threshold, o_activity, d_activity):
    o_zone, o_loc, d_zone, d_loc = self.sequence_stops(stops=stops, etc)
    # code that loops through adding the activities and legs

File: tests/test_9_samplers_tour.py [hunk 10]
```diff
@@ -0,0 +1,138 @@
...
+from pam.core import Person
+from pam.samplers import tour
+from pam.samplers.facility import FacilitySampler
+from pam.variables import END_OF_DAY
+
+# Test input data
+facility_df = pd.DataFrame({'id':[1,2,3,4], 'activity': ['home','delivery','home','depot']})
+points = [Point((1,1)), Point((1,1)), Point((3,3)), Point((3,3))]
+facility_gdf = gp.GeoDataFrame(facility_df, geometry=points)
+
```
fredshone, line 17: nice

tests/test_9_samplers_tour.py
```diff
@@ -0,0 +1,138 @@
...
+def test_distribution_type():
+    bins = range(0,24)
+    pivots = {1: 0, 2: 1, 8: 2, 15: 3}
+    total = 100
+
+    dist = tour.PivotDistributionSampler(bins=bins, pivots=pivots, total=total)
+    print(dist.demand)
+    
+    assert isinstance(dist.demand, dict)
```
fredshone, line 68: make a proposer assertion... i'm actually unsure what happens if the first possible pivot (0) is not defined?

```
    # uniform
    bins = range(0,3)
    pivots = {0:1, 2: 1}
    total = 30
    dist = tour.PivotDistributionSampler(bins=bins, pivots=pivots, total=total)
    assert dist.dict[0] == 10
    assert dist.dict[1] == 10
    assert dist.dict[2] == 10
```

```
    # triangle
    bins = range(0,3)
    pivots = {0:0, 2: 1}
    total = 30
    dist = tour.PivotDistributionSampler(bins=bins, pivots=pivots, total=total)
    assert dist.dict[0] == 0
    assert dist.dict[1] == 10
    assert dist.dict[2] == 20
```

```
    # incomplete
    bins = range(0,3)
    pivots = {1:1, 2:1}
    total = 30
    dist = tour.PivotDistributionSampler(bins=bins, pivots=pivots, total=total)
    assert dist.dict[0] == ?
    assert dist.dict[1] == ?
    assert dist.dict[2] == ?
```

```
    # empty
    bins = range(0,3)
    pivots = {}
    total = 30
    dist = tour.PivotDistributionSampler(bins=bins, pivots=pivots, total=total)
    assert dist.dict[0] == ?
    assert dist.dict[1] == ?
    assert dist.dict[2] == ?
```

tests/test_9_samplers_tour.py
```diff
@@ -0,0 +1,138 @@
...
+    assert len(o_density.density) > 0
+    assert len(d_density.density) > 0
+
+def test_dzone_sampler_dzone_d_density_zero():
+    d_density = tour.FacilityDensitySampler(facility_zone=facility_zone, zone=zones_gdf, activity = 'delivery')
+    o_zone = 3
+    dist_threshold = df_od.median().agg('median')
+
+    with pytest.raises(UserWarning, match='No destinations within this distance'):
+        d_zone = d_density.dzone_sample(o_zone=o_zone, df_od=df_od, dist_threshold=dist_threshold)
```
fredshone, line 88: nice

tests/test_9_samplers_tour.py
```diff
@@ -0,0 +1,138 @@
...
+    o_zone, o_loc, d_zone, d_loc = tour.TourPlan().sequence_stops(stops=stops,
+                                                                ozone_sampler=o_density,
+                                                                dzone_sampler=d_density,
+                                                                df_od=df_od,
+                                                                dist_threshold=dist_threshold,
+                                                                dist_id='distance',
+                                                                facility_sampler=facility_sampler,
+                                                                o_activity='depot',
+                                                                d_activity='delivery')
+    assert print(len(d_loc)) == print(stops)
```
fredshone, line 105: get rid of the print statements

fredshone (changes requested):
it's looking good, but:

(i) comments on `TourPlan(er)` - this can definately be given a bit more structure and remove more burden from the notebooks with an "apply_to(agent)" method.

(ii) before pushing, go through code, check/add/remove docstrings/line spacing/print statements etc

pam/samplers/tour.py
divyasharma-arup: Yes, this was silly. I actually didn't know how to sort a dictionary, which is why I converted to a dataframe. I've definitely googled it, but I should probably spend some time figuring out OrderedDict or something similar. If we know how, welcome any suggestions!! @Theodore-Chatziioannou

fredshone: based on https://stackoverflow.com/questions/613183/how-do-i-sort-a-dictionary-by-value:

```
d_seq = sorted(d_seq, key=lambda item: item.get("distance"))
```
will return a list of the dictionaries sorted by the distance value of each dict.

`lambda item: item.get("distance")` is a bit nasty looking but `lambda` is simply special python function declaration. This one days that the sort key is `item.get("distance")`.

pam/samplers/tour.py
divyasharma-arup: Thanks, this was a good catch. I've updated the code to ensure d_zone is a list of zones that align with the d_locs.

tests/test_9_samplers_tour.py
divyasharma-arup: These assertions have been added.

divyasharma-arup:
Hi @fredshone , I've pushed an updated version of the tour module that captures your edits above and has rearranged a few parts after using the module for our current project. Changes as below:

- Merged FacilityDensitySampler logic with FrequencySampler. With TE, I was not using the density information, but found I still needed functionality to sample a destination based on distance. This has now been added to FrequencySampler. 
- Created a separate "create_density_gdf" method that can still be used with FrequencySampler to sample based on density of locations (this is demonstrated in the 17_tour.ipynb)
- Added a regression line to the validation plots at the end.
- Removed any reference to internal projects within the 17_tour.ipynb notebook.

Please kindly let me know your comments.

fredshone (approved):
great stuff well done for persevering. The example notebook is much slicker i think this is going to make future improvements a bit slicker so let's get this merged.

I approve but please make one final pass of the example notebook:
- rename to something informative like "sampling_tour_based_plans"
- clean out unnecesary white space and comments
- check it runs in order one final time (use run all)

🥳

divyasharma-arup:
@fredshone, cleaned up notebook has been renamed and pushed.

fredshone (approved):
looks great, lgtm! (let's get this merged)

PR #142, opened 2022-05-05
Add a daily scheduled CI build

mfitz:
A minimal daily scheduled CI build that runs the unit tests in an attempt to uncover any problems caused by having loosened our version constraints on some of our dependencies.

See https://arupdigital.atlassian.net/browse/LAB-1639.

Note that, for security reasons, new Actions workflows do not run until they have been created in the default branch, so we won't see this scheduled build start to run until after this PR has been merged.

### DIFF
--- a/.github/workflows/daily-scheduled-ci.yml
+++ b/.github/workflows/daily-scheduled-ci.yml
@@ -59,7 +59,7 @@ jobs:
       env:
         SLACK_COLOR: '#FF0000'
         SLACK_LINK_NAMES: true
-        SLACK_MESSAGE: @here ${{ github.repository }} Daily scheduled CI Build ${{ github.run_number }} has failed
+        SLACK_MESSAGE: '<!here> ${{ github.repository }} Daily scheduled CI Build ${{ github.run_number }} has failed'
         SLACK_TITLE: Daily CI Build Failure!
         SLACK_CHANNEL: pam-feeds
         SLACK_USERNAME: GitHub Build Bot
--- a/.github/workflows/daily-scheduled-ci.yml
+++ b/.github/workflows/daily-scheduled-ci.yml
@@ -1,0 +1,66 @@
+name: Daily PAM CI Build
+
+on:
+  schedule:
+    - cron: '23 14 * * 1-5'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+
+    - uses: actions/cache@v1
+      with:
+        path: ~/.cache/pip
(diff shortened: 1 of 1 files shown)


.github/workflows/daily-scheduled-ci.yml
```diff
@@ -0,0 +1,66 @@
...
+        export CURL_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt
+        ./scripts/code-coverage.sh
+
+    - name: Send build success notification
+      if: success()
+      uses: rtCamp/action-slack-notify@v2.0.0
+      env:
+        SLACK_MESSAGE: ${{ github.repository }} Daily scheduled CI Build ${{ github.run_number }} has succeeded
+        SLACK_TITLE: Build Success
+        SLACK_CHANNEL: city-modelling-feeds
```
mfitz, line 51: Decided on this rather than `pam-feeds` here, but I am open to changing it

.github/workflows/daily-scheduled-ci.yml
```diff
@@ -0,0 +1,66 @@
+name: Daily PAM CI Build
+
+on:
+  schedule:
+    - cron: '23 14 * * 1-5'
```
mfitz, line 5: 2.23 p.m UTC. If you kick off your scheduled builds on the hour like most people, you will sometimes find there is no runner available and your job waits in a queue, so this is an attempt to use an unpopular kick off time.

fredshone (commented):
I think it makes more sense to use the pam-feeds, but i don't feel strongly.

How about a fail message that includes an @here or @akay-arup? To ensure we take notice.

.github/workflows/daily-scheduled-ci.yml
fredshone: how about @akay-arup in the message?

mfitz:
> `unsubscribe`

🤣

.github/workflows/daily-scheduled-ci.yml
```diff
@@ -0,0 +1,67 @@
...
+        SLACK_ICON: https://slack-files2.s3-us-west-2.amazonaws.com/avatars/2017-12-19/288981919427_f45f04edd92902a96859_512.png
+        SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+
+    - name: Send build failure notification
+      if: failure()
+      uses: rtCamp/action-slack-notify@v2.0.0
+      env:
+        SLACK_COLOR: '#FF0000'
+        SLACK_LINK_NAMES: true
+        SLACK_MESSAGE: '<!here> ${{ github.repository }} Daily scheduled CI Build ${{ github.run_number }} has failed'
```
mfitz, line 62: This becomes `@here` after passing through the Slack API

fredshone (approved):

PR #143, opened 2022-05-09
refactor qa checks and add example hook

fredshone:
adds some file size quality assurance to the build pipeline. Additionally refactors scripts to provide some more consistency with matesto repo.

adds a `scripts/code-qa/check-files-sizes.sh`. This is hardcoded to warn about file sizes over 1mb and fail over 10mb.

This script is used in 3 places:
1. added to a "run all" script `scripts/code-qa/qa-checks.sh`
2. added to the example git commit hook
3. added to the build pipeline

I am unsure if we want to enforce this in the build?

Main focus should be firstly showing users the script and then using the hook. Such that we nip large commits in the bud before they can make it into the history.

Also worth noting that commits are now quite time expensive (have to wait for tests and smoke tests) unless using `--no-verify`.


scripts/code-qa/qa-checks.sh
```diff
@@ -0,0 +1,17 @@
...
+# fail for size that is too large
+
+
+set -e
+
+pushd "${0%/*}"
+
+./code-coverage.sh
+./notebooks-smoke-test.sh
+./check-file-sizes.sh
```
mfitz, line 15: I guess it's a belt-and-braces thing, but I definitely want to know about the file sizes at the point I'm trying to **commit**, but I want to perform the other checks in this script a lot more frequently than that, regardless of the state in git. The file size check definitely belongs in the commit hook, but possibly not here. That said, I don't suppose it does any harm.

mfitz (changes requested):
Looks good, just a few minor suggestions in the comments.

We should also have as forceful an instruction as we can muster in the README about installing the Git hook. It should be part of the "installing" section. I am not aware of a way to automate installation of git hooks, say on cloning the repo; allowing a repo to run arbitrary code on any machine it is cloned to would be an enormous security problem. But perhaps we could wrap up the `pip install` and the installation of the git hook into a single `install.sh` and instruct the user to run the script? We'd need an equivalent `install.bat` file for windows users too, I guess? At the very least, the README should push the user to install the commit hook.

.github/workflows/build_pipeline.yml
```diff
@@ -37,10 +37,13 @@ jobs:
     - name: Run tests
       run: |
         export CURL_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt
-        ./scripts/code-coverage.sh
+        ./scripts/code-qa/code-coverage.sh
     - name: Smoke test notebooks
       run: |
-        ./scripts/notebooks-smoke-test.sh
+        ./scripts/code-qa/notebooks-smoke-test.sh
+    - name: Check for large files
```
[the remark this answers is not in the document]
fredshone: of course 🥇 - I will remove.

But getting the warning is useful  - in the short term to warn us BFG might be necessary, and longer term failed builds are painful enough that people will eventually learn.

I will propose a script that checks them all.

Do you think if a file is then removed, that squash commit (and delete old branch) might keep it out of history?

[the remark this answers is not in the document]
mfitz: I think that deleting an unmerged branch probably doesn't remove its commits from history, but they would be dangling references that might eventually be removed by garbage collection on the repo. I'll double check that though.

scripts/code-qa/check-file-sizes.sh
```diff
@@ -0,0 +1,41 @@
...
+		echo "FAILURE $file is over $TOO_LARGE KB size ($file_size KB)."
+		((HAS_ERROR++))
+	elif [ "$file_size" -ge "$WARNING_SIZE" ]; then
+		echo "WARNING $file is over $WARNING_SIZE KB size ($file_size KB)."
+		((HAS_WARNING++))
+	fi
+done
+
+if [ "$HAS_ERROR" != 0 ]; then
+	echo "FAILURE: $HAS_ERROR files are over $TOO_LARGE KB."
```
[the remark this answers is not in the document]
fredshone: yes i had the same thought but I'm currently planning to enforce file sizes in the build as you say

fredshone:
thanks @mfitz - I think i've incorporated all your comments.

I've created a check "all" script for file sizes that i am trying out in the build. It alrady identified a chonky example dataset which i have reduced.

README.md
```diff
@@ -89,32 +89,50 @@ cd pam
...
+
+- `scripts/code-qa/qa-checks.sh` - run a sensible combination of all the following
+- `pytest` - run the test suite
+- `scripts/code-qa/code-coverage.sh` - run tests and check code coverage
+- `scripts/code-qa/notebooks-smoke-test.sh` - smoke test all notebooks found in the `examples` directory
+- `scripts/code-qa/check-staged-file-sizes.sh` - check for large file sizes that have been git staged
+- `scripts/code-qa/check-all-file-sizes.sh` - check for large file sizes
+
+**Git Commit Hook** </br>
+You can automate the above checks using our git commit hook. To implement this automation simply copy `scripts/got-hooks/pre-commit` to `.git/hooks/pre-commit`. eg for OSX:
```
mfitz, line 103: `got-hooks`? 😃

README.md
```diff
@@ -65,8 +65,8 @@ source venv/bin/activate
 pip install -e .
 ```
 
-We known that requirements are already available in a up to date Anaconda installation. 
-If you're not using Anaconda, sometimes the above will fail due to spatial library dependencies not being properly installed using pip. If you see any errors relating to fiona or rtree this is likely the case. To fix it, wipe the virtual environment and recreate, `brew install spatialindex` (Mac) then install everything else in the `venv` using `pip install -e .` again. 
+We known that requirements are already available in a up to date Anaconda installation.
```
mfitz, line 68: I know this is not new in this change, but there is a typo here that we should fix:

> We known that...

scripts/code-qa/check-all-file-sizes.sh
```diff
@@ -0,0 +1,48 @@
...
+
+echo "Starting check of all project file sizes."
+echo "File size warning threshold is set to $WARNING_SIZE_KB KB"
+echo "File size fail threshold is set to $TOO_LARGE_KB KB"
+echo "-----------"
+
+pushd "${0%/*}/../.." > /dev/null
+
+HAS_WARNING=0
+HAS_ERROR=0
```
mfitz, line 20: Slightly better names would be `NUMBER_OF_WARNINGS` and `NUMBER_OF_ERRORS`, because they are both essentially counters

scripts/code-qa/check-all-file-sizes.sh
```diff
@@ -0,0 +1,48 @@
...
+
+WARNING_SIZE_KB=1000 #KB
+TOO_LARGE_KB=10000 #KB
+
+echo "Starting check of all project file sizes."
+echo "File size warning threshold is set to $WARNING_SIZE_KB KB"
+echo "File size fail threshold is set to $TOO_LARGE_KB KB"
+echo "-----------"
+
+pushd "${0%/*}/../.." > /dev/null
```
mfitz, line 17: Nice.

mfitz (changes requested):
Couple of picky naming comments and some minor typos to fix, one of which is in the new documentation, so wants correcting.

scripts/code-qa/qa-checks.sh
```diff
@@ -0,0 +1,17 @@
...
+# fail for size that is too large
+
+
+set -e
+
+pushd "${0%/*}"
+
+./code-coverage.sh
+./notebooks-smoke-test.sh
+./check-staged-file-sizes.sh
```
mfitz, line 15: This has only just occurred to me, but I think the linting check should be a script in its own right ([like this one](https://github.com/arup-group/matesto/blob/main/scripts/code-qa/lint-check.sh)) which we invoke from this script, [as we do in other projects](https://github.com/arup-group/matesto/blob/main/scripts/code-qa/qa-checks.sh). I see that in the CI build we are restricting the linting rules in play (`--select=E9,F63,F7,F82`) - I don't think we should do that. 

When I run linting using the full set of rules, there are 767 errors, so probably we should handle that in a separate PR. I will raise an issue for it and address it soon:

```bash
flake8 . --max-line-length 120 --count --show-source --statistics --exclude=scripts,tests
...
2     E101 indentation contains mixed spaces and tabs
12    E122 continuation line missing indentation or outdented
3     E124 closing bracket does not match visual indentation
33    E125 continuation line with same indent as next logical line
4     E128 continuation line under-indented for visual indent
61    E203 whitespace before ':'
4     E221 multiple spaces before operator
20    E225 missing whitespace around operator
1     E227 missing whitespace around bitwise or shift operator
1     E228 missing whitespace around modulo operator
95    E231 missing whitespace after ':'
271   E251 unexpected spaces around keyword / parameter equals
21    E261 at least two spaces before inline comment
6     E262 inline comment should start with '# '
9     E265 block comment should start with '# '
13    E266 too many leading '#' for block comment
1     E271 multiple spaces after keyword
22    E302 expected 2 blank lines, found 1
10    E303 too many blank lines (2)
52    E501 line too long (150 > 120 characters)
1     E502 the backslash is redundant between brackets
12    E711 comparison to None should be 'if cond is not None:'
1     E712 comparison to True should be 'if cond is True:' or 'if cond:'
3     E713 test for membership should be 'not in'
22    E999 SyntaxError: invalid syntax
5     F401 'pam.plot.plans.*' imported but unused
5     F403 'from pam.plot.plans import *' used; unable to detect undefined names
2     W191 indentation contains tabs
36    W291 trailing whitespace
4     W292 no newline at end of file
34    W293 blank line contains whitespace
1     W391 blank line at end of file
767
```

mfitz (approved):

PR #146, opened 2022-05-26
V12 io default

fredshone:
moves matsim read write functions to default to "version 12"


Theodore-Chatziioannou (approved):
great. 
Were those tests breaking with v12? If so, perhaps we should follow up with a separate PR with appropriate v12+ tests.

fredshone:
Coverage should be ok - the tests changed were v11 tests but were relying on the default v11 behaviour. We have additional coverage for v12 already i think.

PR #147, opened 2022-06-06
implements one-hot encode decode and distances

fredshone:
no one really needs this PR yet but i think it's kind of cool:

mostly implements a function to transform a Plan into a one-hot encoded array:
```
from pam import array
 encoded = array.encode.plan_to_one_hot(
  plan=plan,
  mapping={"home":0, "work":1, "travel":2},
  bin_size=3600,  # seconds
  duration=86400  # seconds
  )
```
This proves a 2d numpy array with the first dimension representing time bins and the second a one-hot encoding of activity and travel, eg:
```
[[1,0,0],[1,0,0],[0,1,0],[0,0,1]...]  # home-home-travel-work...
```
This is designed to work with the VAE experiments (although this returns only the array for a single plan and is only 2d, whereas existing experiment have a third dimension (which is redundant)).

also provides a "decode" function, such that plans can be (mostly) recovered:
```
encoded_plan = array.encode.plan_to_one_hot(
  plan = plan,
  mapping = {"home":0, "work":1, "travel":2},
  bin_size=3600,  # seconds
  duration=86400  # seconds
    )
decoded_plan = array.decode.one_hot_to_plan(
    array = encoded_plan,
    mapping = {0:"home", 1:"work", 2:"travel"},
  bin_size=3600,  # seconds
  duration=86400  # seconds
)
```
Some info is obviously lost. For example there is no encoding of locations.

Activity and Leg times and durations are also binned such that there is some precision loss and short components can be lost altogther (ie activities or trips shorter than the bin size). The decode method will go some way to recover a valid plan sequence by inferring missing start and end activities, and ensuring a proper sequence. However locations are not recovered (something to think about in future).

Other than supporting the VAE work. This is also buit to provide a simple way of comparing activity sequences:

```
encoded_plan_A = array.encode.plan_to_one_hot(
        plan = plan_A,
        mapping = {"home":0, "work":1, "travel":2},
    )
encoded_plan_B = array.encode.plan_to_one_hot(
        plan = plan_B,
        mapping = {"home":0, "work":1, "travel":2},
    )

print(array.distance.accuracy(encoded_plan_A, encoded_plan_B)
print(array.distance.test_one_hot_cross_entropy(encoded_plan_A, encoded_plan_B)
```

These are pretty rudimentary measures of difference but i fancy them for some other experiment i am going to try so here they are.

### DIFF
--- a/pam/array/decode.py
+++ b/pam/array/decode.py
@@ -1,0 +1,139 @@
+import numpy as np
+from datetime import datetime, timedelta
+
+from pam.activity import Plan, Activity, Leg
+from pam.variables import START_OF_DAY, END_OF_DAY
+
+
+def one_hot_to_plan(
+    array:np.array,
+    mapping:dict,
+    bin_size:int=3600,
+    duration:int=86400,
+    start_of_day:datetime=START_OF_DAY,
+    end_of_day:datetime=END_OF_DAY,
+    leg_encoding:str="travel",
+    default_leg_mode:str="car",
+    default_activity:str="other"
+    ) -> Plan:
+    """
+    Decode a one-hot encoded plan array for a given mapping. Attempts to create a valida plan sequence
+    by assuming obviously missing components. Does not support locations, these must be created
+    manually.
+
+    Args:
+        array (np.array): input one-hot encoded plan.
+        mapping (dict): encoding index to activity, eg {0: 'home', 1:'travel'}.
+        bin_size (int, optional): in seconds. Defaults to 3600.
+        duration (int, optional): in seconds. Defaults to 86400.
+        start_of_day (datetime, optional): start datetime of first activity. Defaults to START_OF_DAY.
+        end_of_day (datetime, optional): end time of last activity. Defaults to END_OF_DAY.
+        leg_encoding (str, optional): activity encoding for travel components. Defaults to "travel".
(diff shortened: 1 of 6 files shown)


pam/array/decode.py
```diff
@@ -0,0 +1,139 @@
...
+    bin_size:int=3600,
+    duration:int=86400,
+    start_of_day:datetime=START_OF_DAY,
+    end_of_day:datetime=END_OF_DAY,
+    leg_encoding:str="travel",
+    default_leg_mode:str="car",
+    default_activity:str="other"
+    ) -> Plan:
+    """
+    Decode a one-hot encoded plan array for a given mapping. Attempts to create a valida plan sequence
```
alex-kaye, line 20: typo ("valid")

pam/array/decode.py
```diff
@@ -0,0 +1,139 @@
...
+        start_of_day=start_of_day,
+        bin_size=bin_size
+        ):
+        if act == leg_encoding:  # add leg
+            proposed_plan.append(Leg(mode=default_leg_mode, start_time=start_time))
+        else:
+            proposed_plan.append(Activity(act=act, start_time=start_time))
+
+    add_end_times(proposed_plan, end_of_day)
+    print(proposed_plan)
```
alex-kaye, line 58: remove print statement

alex-kaye (approved):
minor comments added

PR #151, opened 2022-06-27
Plan cropping

Theodore-Chatziioannou:
Implements the plan cropping methods as a PAM module (incl some tests). 
A CLI module is also introduced, which can invoke the cropping tool.


pam/cli.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    Crop a population's plans outside a core area.
+    :param path_population_input: Path to a MATSim population (xml)
+    :param path_boundary: Path to the core area geojson file
+    :param dir_population_output: Path to the output (cropped) MATSim population 
+    :param matsim_version: MATSim version
+    :param comment: A short comment included in the output population
+
+    """
+    from pam.cropping import crop_xml
```
fredshone, line 31: prefer imports at top. I think the main reason to have in a function def is to either resolve some circular dependancy or avoid some expensive external import?

fredshone (approved):
great - i approve, but please consider:
i) moving the mid file import
ii) adding a buffer command
buffering seems like a cheap but useful feature. A zero buffer is also a great way of fixing broken geoms.

pam/cli.py
```diff
@@ -0,0 +1,47 @@
...
+    pass
+
+
+@cli.command()
+@click.argument("path_population_input", type=click.Path(exists=True))
+@click.argument("path_boundary", type=click.Path(exists=True))
+@click.argument("dir_population_output", type=click.Path(exists=False, writable=True))
+@click.option("--matsim_version", "-v", default=12)
+@click.option("--comment", "-c", default="cropped population")
+def crop(path_population_input, path_boundary, dir_population_output,
```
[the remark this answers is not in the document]
Theodore-Chatziioannou: added

pam/cli.py
Theodore-Chatziioannou: yeah, it was mainly for performance reasons, thinking we may add CLI support for multiple modules in the future. But I agree - moved it at top

PR #152, opened 2022-06-29
sample cli

Theodore-Chatziioannou:
Allows down- or up-sampling a MATSim (xml) population via the command line, by invoking: `pam sample <path_population_xml> <path_core_area_geojson> <path_output_directory> -s <sample_percentage> -v <matsim_version>` 

For example, we can decrease the size of a hypothetical `plans.xml` file from 1000 agents to 100 by running `pam sample plans.xml output/resampled -s 0.1`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -614,5 +614,6 @@ The `pam.cropping` module allows to spatially subset populations, by simplifying
 Some of PAM's functionality can be invoked via its Command Line Interface (CLI). The currently supported methods are: 
 
 * plan cropping: `pam crop <path_population_xml> <path_core_area_geojson> <path_output_directory>`.
+* down/up-sampling an xml population: `pam sample <path_population_xml> <path_core_area_geojson> <path_output_directory> -s <sample_percentage> -v <matsim_version>`. For example, you can use: `pam sample tests/test_data/test_matsim_plansv12.xml tests/test_data/output/sampled -s 0.1` to create a downsampled (to 10%) version of the input (`test_matsim_plansv12.xml`) population.
 
 To further explore the CLI commands, start by typing `pam --help` in a terminal.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -614,5 +614,6 @@ The `pam.cropping` module allows to spatially subset populations, by simplifying
 Some of PAM's functionality can be invoked via its Command Line Interface (CLI). The currently supported methods are: 
 
 * plan cropping: `pam crop <path_population_xml> <path_core_area_geojson> <path_output_directory>`.
+* down/up-sampling an xml population: `pam sample <path_population_xml> <path_core_area_geojson> <path_output_directory> -s <sample_percentage> -v <matsim_version>`. For example, you can use: `pam sample tests/test_data/test_matsim_plansv12.xml tests/test_data/output/sampled -s 0.1` to create a downsampled (to 10%) version of the input (`test_matsim_plansv12.xml`) population.
```
alex-kaye, line 617: - `<path_core_area_geojson>` should be removed for the `pam sample` function.

- Since `test_matsim_plansv12.xml` only contains 5 plans, it may be worth sampling at least 20% in the example statement so that it returns one full plan.  i.e. `pam sample tests/test_data/test_matsim_plansv12.xml tests/test_data/output/sampled -s 0.2`

fredshone: nice high quality reviewing!

fredshone (approved):
happy once AKs comments are addressed

README.md
Theodore-Chatziioannou: updated the readme. 
I haven't added a "downsampling" test because there a small element of randomness. During downsampling, every agent has 20% chance of being sampled, but this will not always result in a population that is exactly 20% in size (in the vast majority of cases it will, though). This would create a risk of a test occasionally breaking for no apparent reason in the future. 
However, we should indeed test for more cases. I have parameterized the test to include a scaling factor of 100%, in which case we should always get the initial population.

PR #154, opened 2022-07-07
adds summary report

fredshone:
adds `report` group to cli with `summary` command. Intention is to provide standard output for describing a MATSim population/plans. The primary driver for this is to improve handovers.

Also some general core and cli fiddling but nothing significant.

```
❯ pam report summary --help
Usage: pam report summary [OPTIONS] PATH_POPULATION_INPUT

  Summarise a population.

  :param path_population_input: Path to a MATSim population (xml) :param
  sample_size: The sample size.     For example, use 0.1 to produce a 10%
  version of the input population :param matsim_version: MATSim version :param
  attribute_key: A random seed. :param household_key: Household key, defaults
  to 'hid'.

Options:
  -s, --sample_size FLOAT
  -v, --matsim_version INTEGER
  -k, --attribute_key TEXT
  -h, --household_key TEXT
  --help                        Show this message and exit.
```

image

Implementation is super slow, requires loading MATSim population to memory then looping through population multiple times. But good enough for now.


Theodore-Chatziioannou (changes requested):
this is really useful @fredshone .

A couple of requests:
- could you add a bullet in the CLI section of the readme? With a simple example that a user can run (ie `pam report summary tests/test_data/test_matsim_plansv12.xml`).
- could you add some explanation of what the `attribute_key` argument is for? I was able to break the cli with it in some experiments, so it may need a safety net.

Theodore-Chatziioannou:
could you add an option for saving to a text file, so that the results can be more easily shared?
I was hoping that we could simply use `pam report summary tests/test_data/test_matsim_plansv12.xml > log.txt`, but I think the pretty-print is causing some glitches:
image

fredshone:
Added a simple section to the readme (e can probably move a lot of the policy stuff out now).
The segmentation option `-k` is documented by `--help` and should be better behaved now.
Moved cli doc string info into help such that appears when exploring the cli:
```
Usage: pam report summary [OPTIONS] PATH_POPULATION_INPUT

  Summarise a population.

Options:
  -v, --matsim_version INTEGER  MATSim plan format, default 12.
  -d, --debug                   Switch on debug verbosity.
  -s, --sample_size FLOAT       Input sample size. Default 1. For example, use
                                0.1 to apply a 10% weighting to the input
                                population.
  -k, --attribute_key TEXT      Optional population attribute key to segment
                                output, eg 'subpopulation'.
  -h, --household_key TEXT      Household key, default 'hid'.
  --crop / --no-crop            Crop or don't crop plans to 24 hours, defaults
                                to crop.
  --rich / --text               Formatted (for terminal) or regular text
                                output (for .txt).
  --help                        Show this message and exit.
```
The weird formatting is caused by the colour formatting, for now i've added a `--text` option that provides a simple text output.  Not the nicest approach but fine for now. The colour stuff seems very superfluous but it does make it a lot easier to read... 🤷🏻

PR #155, opened 2022-07-11
Plans stream aka "print plans to terminal"

fredshone:
This implements some viz used for the pam Arup University video, but also implements some useful patterns.

Alex, Divya, Jose, the more eyes the merrier for PRs - but no pressure to review in detail.

1. Adds terminal plan "plot" to cli:
```
pam report stringify --help
Usage: pam report stringify [OPTIONS] PATH_POPULATION_INPUT

  ASCII plot activity plans to terminal.

Options:
  --colour / --bw      Choose a colour or grey-scale (bw) output, default
                       'colour'
  -w, --width INTEGER  Target character width for plot, default 72
  --help               Show this message and exit.
```

eg:

`pam report stringify tests/test_data/test_matsim_plansv12.xml`
Screenshot 2022-07-11 at 14 40 12

2. MATSim read stream:

Includes matsim read function for incremental reading aka 'stream': `pam.read.stream_matsim_persons`. This can be used to iterate through persons in a plans file. If we implement a comparable write method we can implement a lot of the cli methods without heavy memory requirements, such that they can be run on a small box.

This is only useful if the person plans are be modified without needing to know the whole household or population (perhaps household if we can guarantee occupants are consecutive in the xml). So spatial cropping, simple sampling, jitter, potentially optimisation, simple policies etc would be fine as streamed, but not facility sampling or pam.planner.

I will refactor the MATSim read functions to have less duplication of code. Also I am thinking to make this only support v12+.

3. Categorical encoder:

Includes a categorical encoder `pam.array.encode.PlansToCategorical`. This is implemented as a class so that it is persistent and can build a mapping between activity names as the categorical encoding 'on the fly'. I think the one-hot encoder should also be changed to have this pattern.


pam/read.py
```diff
@@ -890,7 +890,66 @@ def read_matsim(
...
+    Does not support pre version 12 attributes inputs.
+    :param plans: path to matsim format xml
+    :param attributes: path to matsim format xml
+    :param all_vehicles_path: path to matsim all_vehicles xml file
+    :param electric_vehicles_path: path to matsim electric_vehicles xml
+    :param weight: int
+    :param version: int {11,12}, default = 12
+    :param household_key: {str, None}
+    :param keep_non_selected: Whether to parse non-selected plans (storing them in person.plans_non_selected).
+    :return: Population
```
Theodore-Chatziioannou, line 913: :return: pam.core.Person

Theodore-Chatziioannou (approved):

PR #156, opened 2022-07-14
Tour module

divyasharma-arup:
The tour module has been updated as below:

**Tour Module:**

1. The TourPlanner has been updated to accept an origin zone input instead of origin zone sampler. 
This enables the user to create a destination matrix that is dependent on the origin zone, which may be important for samplers based on o/d flows. It can still be used as is for cases where the destination is not dependent on the origin zone (i.e, facility density). 

2. The TourPlanner now has logic to sample if there is no threshold_matrix input. 
This enables the user to decide whether the destination zone will be based on a threshold value or randomly sampled from the input destination distribution. 

**Example Notebok:**
Removed visualisations to reduce the size of the notebook.

**Tour Module Tests:**

Updated tests to reflect the TourPlanner changes. Also, I attempted to refactor the tests of the Tour Planner so that the code is a little less repetitive. The agent and agent plan are not defined in each individual test case, but instead copied to an object agent_test, agent_plan_test for each test case. Uncertain if this is the best way/naming convention to minimise the repetiveness in the code.


fredshone (approved):
Nice job bringing this all together.

I added a few tests for my own assurance. All good.

There are a few slightly odd patterns which we will iron out one day. But i am extremely impressed everything working so well - approved -let's get this merged!

PR #157, opened 2022-07-14
Population combiner

alex-kaye:
This combines multiple populations via the pam cli. 

I would like to add a validation check that avoids existing `combined_population.xml` files being overwritten, if this file already exists. My attempt at adding this is `cli.py line 180`. 

Please could you review and add any improvements.


pam/cli.py
```diff
@@ -176,6 +177,61 @@ def crop(
     logger.info(f'Output saved at {dir_population_output}/plan.xml')
 
 
+# avoid_override() is not wokring correctly - expecting a prompt to appear only when the outpath already exists
+def avoid_override(
+    population_output: str
+    ):
+    if os.path.exists(population_output) == True:
+        population_output.abort()
```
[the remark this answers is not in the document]
fredshone: (not tested)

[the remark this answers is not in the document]
fredshone: ok now i read the docs... doesn't look like what you want is going to obviously work as a call-back. Seems it is hard-coded to always ask y/n regardless of exists or not.

I think you will need to add you own logic at the start of the command (after receiving the args). Something like:

```
def callback_if_exists(path):
    if os.path.exists(path) == True:
       if input(f"{population_output} exists, overwrite? [y/n]:") .lower() in ["y", "yes", "ok"]:
           return None
       else:
           raise UserWarning(f"Aborting to avoid overwrite of {population_output}")
```

pam/cli.py
```diff
@@ -176,6 +177,61 @@ def crop(
...
+
+@cli.command()
+@common_options
+@click.argument(
+    "population_paths", type=click.Path(exists=True), nargs=-1
+    )
+@click.option(
+    "--population_output", "-o", type=click.Path(exists=False, writable=True), default=os.getcwd()+"\combined_population.xml",
```
fredshone, line 193: I think just `default="combined_population.xml"` will work fine?

pam/cli.py
```diff
@@ -176,6 +177,61 @@ def crop(
...
+
+    write.write_matsim(
+        population = combined_population,
+        version=matsim_version,
+        plans_path=population_output,
+        comment=comment
+    )
+    logger.info('Population combiner complete')
+    logger.info(f'Output saved at {population_output}')
+
```
fredshone, line 231: 4 blank lines? :)

pam/operations/population_combiner.py
```diff
@@ -0,0 +1,29 @@
...
+
+    combined_population = core.Population()
+
+    for inpath in inpaths:
+     
+        population = read.read_matsim(
+            os.path.join(inpath, "plans.xml"),
+            household_key="hid",
+            weight=1,
+            version=12
```
fredshone, line 23: your command accepts a version arg. Users will prob assume it works on both the read and write...

personally don't mind if you hard code it all to 12. But otherwise make accessible here as well.

fredshone (changes requested):
looks good, nice clean pattern, some minor comments and then we need to add a few tests. Let me know once you're done and I'll add some for you. Then we will go through together.

pam/cli.py
alex-kaye: nice! thank you. That works.

I didn't define a `callback_if_exists` function but just used the logic. It seems each cli command only likes to be passed one function.

pam/cli.py
alex-kaye: I keep getting permission errors without defining the cwd, so have kept this in

fredshone (approved):
let's get this merged

alex-kaye:
thank you - were the tests okay/sufficient?

PR #161, opened 2022-07-24
Adds CLI versioning

chicken-teriyaki-cup-rice:
Based on https://click.palletsprojects.com/en/8.1.x/api/.

Before:
before

After:
after

Actual output:
actual_output

### DIFF
--- a/pam/cli.py
+++ b/pam/cli.py
@@ -8,6 +8,7 @@ from pam.cropping import crop_xml
 from pam.samplers import population as population_sampler
 from pam import read, write
 from pam.report.summary import pretty_print_summary, print_summary
+from pam.report.stringify import stringify_plans
 
 
 logging.basicConfig(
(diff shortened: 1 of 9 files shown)


pam/cli.py
```diff
@@ -36,6 +36,7 @@ def cli():
 
 
 @cli.group()
+@click.version_option()
```
mfitz, line 39: I think maybe this is in the wrong place - it should be on line 31 so that it is applied to the root command. As it stands, I can only see the `--version` option on the `pam report` subcommand:

```bash
$ pam --help
Usage: pam [OPTIONS] COMMAND [ARGS]...

  Population Activity Modeller (PAM) Command Line Tool

Options:
  --help  Show this message and exit.

Commands:
  crop    Crop a population's plans outside a core area.
  report  Various reporting for MATSim formatted plans.
  sample  Down- or up-sample a PAM population.
  
$ pam --version
Usage: pam [OPTIONS] COMMAND [ARGS]...
Try 'pam --help' for help.

Error: No such option: --version
```

```bash
$ pam report --help
Usage: pam report [OPTIONS] COMMAND [ARGS]...

  Various reporting for MATSim formatted plans.

Options:
  --version  Show the version and exit.
  --help     Show this message and exit.

Commands:
  summary  Summarise a population.
  
  
$ pam report --version
pam, version 0.0.1
```

mfitz (changes requested):
Needs a minor adjustment to apply the `--version` option to the correct command.

PR #162, opened 2022-07-26
refactors benchmarks into report module

fredshone:
Originally set out to make the "batch output" of benchmarks faster and add to the cli:

```
❯ pam report benchmarks --help
Usage: pam report benchmarks [OPTIONS] POPULATION_INPUT_PATH OUTPUT_DIRECTORY

  Write batch of benchmarks to directory

Options:
  -v, --matsim_version INTEGER  MATSim plan format, default 12.
  -d, --debug                   Switch on debug verbosity.
  --help                        Show this message and exit.
```

This dumps a bunch of benchmarks to disk and crucially only iterates through population once. But also:

- Adds a trip_df method, so that in the case of extracting benchmarks from an "experienced" MATSim plan - legs are ignored
- adds a mode share bm
- BREAKING refactors individual bms such that they no longer can write to disk directly - this is personal preference to simplify their interface, reduce function scope for testing and push such use to the cli in future
- BREAKING moves to reporting.benchmarks module and renames benchmarks - in general to remove duplication of the word benchmarks, ie `write.mode_counts_benchmark()` -> `report.benchmarks.mode_counts()`


fredshone:
I've also added "spinners" to the cli. Including a slight refactor of `pam.crop` to expose what is going on to the cli module.

fredshone:
i will resolve conflicts

PR #163, opened 2022-08-15
Write leg attributes

fredshone:
Pretty nasty refactor to try and sensibly allow pam to read/write "warm" plans. Where warm starting is where we use output plans from one run as input to the next. But specifically in this case we want to read/write and potentially modify the intermediate plans using pam.

Specifically pam:
- No longer defaults to "crop" input MATSim plans (cropping of everything after 24 hours)
- NEEDS to read the "routing mode" leg attributes (otherwise warm start fails)
- WANTS to read the leg "route" which includes the network or transit route as appropriate
- Still supports v11 (sigh)
- this does not include the non-selected plans (but i may add this)

I have added the required read/write operations into pam.read and pam.write. I also did some refactoring of the modules to reduce module length. Imports are all the same though - so no breaking changes from this.

_______________

@alex-kaye @panostsolerid in Theo's absence I do require a somewhat careful review by the two of you. I have good test coverage for all my changes so I am not so worried about it works but more so about style, understanding etc. @gac55 - see applications below:

Please note that these are very MATSim centric changes - designed to essentially not loose simulation information from one MATSim run to another if we want to make changes to the population between the MATSim runs. They don't apply to synthesis or diary data read/write ops.
_______________

# Application
A simple example application is if we want to correct or change agent attributes mid-way through a project/run. We can stop simulation. Take latest plans. Load into pam. Make programatic changes. Write back to xml and restart simulation, all without loosing behaviour "learnt" from previous run.

A useful thing we might do in future is make changes to some agents' plans such as their sequences or destinations (eg WfH). We can do this to the selected agents while maintaining the others' "learning". Such that we can warm start (saving time and money) even after population changes.

Similarly we might make infrastructure changes or schedule changes to a simulation for a scenario. We can just "wipe the memory" of the affected agents while maintaining the unaffected:

```
#pseudo
for plan in plans:
    for leg in plan.legs:
        if leg.route.is_routed:
            if any(x in leg.route.network_route for x in removed_links):
                leg.attrinutes = {}  # maybe required?
                leg.route = Route() # empty
```

Long term we might also use this as an interface to transfer learning between agents or force exploration of certain behaviours, for example by manually adding or swapping proposed plans into agent "memories".

_______________

# Details:

## Leg attributes:
Leg attributes are stored in memory as a dictionary.

The attribute write has a pretty janky bit of logic to deal with a "special" MATSim java types (also added something similar for special person attribute type "vehicle"):

```
    for k, v in component.attributes.items():
        if k == 'enterVehicleTime':  # todo make something more robust for future 'special' classes
            attribute = et.SubElement(
            attributes, 'attribute', {'class': 'java.lang.Double', 'name': str(k)}
            )
        else:
            attribute = et.SubElement(
                attributes, 'attribute', {'class': 'java.lang.String', 'name': str(k)}
                )
        attribute.text = str(v)
```

This will do for now but we will likely in future have to either:
(i) read/store/write the attribute classes
or (ii) at least make the special treatment of certain attribute names/keys more explicit somehow

## Leg route:
Leg route is stored in memory as a pam.activity.Route object. This simply wraps the input xml element and provides some hopefully useful properties such as `leg.route.is_transit` and "get" operations such as `leg.route.transit.get("boardingTime")`. This did break some interface with other modules such as `pam.scoring` - now fixed.

The high level implication is that a Leg.Route (even an empty one) is always held in memory (maybe bad), but that the required getting and setting is lazy (maybe good) and the Leg interface are now simpler (definitely good).

Another decision is to hold the info as the original xml, such that it can be simply written back to disk using Route.xml. This is convenient and robust to silent changes in the MATSim data, but a different pattern to elsewhere, where we generally unpack all the required information into core classes. Such as `leg.mode: str` and `leg.starting_time: datetime`.

___________________

# Other

1. Refactored `pam.read.matsim` to remove code duplication.
2. Refactored both read and write modules into sub modules, no breaking changes
3. Renamed most of the tests to maintain (I think) a desirable run order.

___________________

# Todo

1. Adding route objects may have nasty memory and speed implications but i cannot get the profiling scripts to work... 
2. Would be nice to add a streaming write method. Will add issue.
3. Should also support the "unselected plans".


pam/read/matsim.py
```diff
@@ -0,0 +1,520 @@
...
+    Either remove the attributes_path arg, or enable version 11.
+    """
+    )
+
+    if version not in [11, 12]:
+        raise UserWarning("Version must be set to 11 or 12.")
+
+    if version == 11 and not attributes_path:
+        logger.warning(
+"""
```
alex-kaye, line 67: indent docstring

pam/read/matsim.py
```diff
@@ -0,0 +1,520 @@
...
+    simplify_pt_trips : bool = False,
+    autocomplete : bool = True,
+    crop : bool = False,
+    keep_non_selected : bool = False,
+    leg_attributes : bool = True,
+    leg_route : bool = True,
+    ) -> core.Person:
+    """
+    Stream a MATSim format population into core.Person objects.
+    Expects agent attributes (an vehicles) to be supplie as optional dictionaries, this allows this
```
alex-kaye, line 131: *and vehicles
*supplied

pam/read/matsim.py
```diff
@@ -0,0 +1,520 @@
...
+    function to support 'version 11' plans.
+    todo: a v12 only method could also stream attributes and woul use less memory
```
alex-kaye, line 133: *would

pam/read/matsim.py
```diff
@@ -0,0 +1,520 @@
...
+    :param plans: path to matsim format xml
+    :param attributes: {}, map of person attributes, only required for v11
+    :param vehicles: {}, map of vehciles
```
alex-kaye, line 136: *vehicles

pam/read/matsim.py
```diff
@@ -0,0 +1,520 @@
...
+    :param electric_vehicles_path: path to matsim electric_vehicles xml
+    :param weight: int
+    :param version: int {11,12}, default = 12
+    :param simplify_pt_trips: bool, simplify legs in multi-leg trips, defaul t= True
```
alex-kaye, line 140: *default

pam/write/diary.py
```diff
@@ -0,0 +1,212 @@
...
+    for hid, hh in population.households.items():
+        hh_data = {
+            'hid': hid,
+            'freq': hh.freq,
+            'hzone': hh.location.area,
+        }
+        if isinstance(hh.attributes, dict):
+            hh_data.update(hh.attributes)
+        # if hh.location.area is not None:
+        #     hh_data['area'] = hh.location.area
```
alex-kaye, line 46: Remove

pam/write/diary.py
```diff
@@ -0,0 +1,212 @@
...
+            people_data = {
+                'pid': pid,
+                'hid': hid,
+                'freq': person.freq,
+                'hzone': hh.location.area,
+            }
+            if isinstance(person.attributes, dict):
+                people_data.update(person.attributes)
+            # if hh.location.area is not None:
+            #     people_data['area'] = hh.location.area
```
alex-kaye, line 62: remove

pam/write/diary.py
```diff
@@ -0,0 +1,212 @@
...
+                        'mode': component.mode,
+                        'seq': component.seq,
+                        'tst': component.start_time,
+                        'tet': component.end_time,
+                        'duration': str(component.duration),
+                    }
+                    # if component.start_location.area is not None:
+                    #     leg_data['start_area'] = component.start_location.area
+                    # if component.end_location.area is not None:
+                    #     leg_data['end_area'] = component.end_location.area
```
alex-kaye, line 88: remove

pam/write/diary.py
```diff
@@ -0,0 +1,212 @@
...
+                        'freq': component.freq,
+                        'activity': component.act,
+                        'seq': component.seq,
+                        'start time': component.start_time,
+                        'end time': component.end_time,
+                        'duration': str(component.duration),
+                        'zone': component.location.area,
+                    }
+                    # if component.location.area is not None:
+                    #     act_data['area'] = component.location.area
```
alex-kaye, line 107: remove

pam/read/matsim.py
```diff
@@ -0,0 +1,520 @@
...
+
+def unpack_leg_v12(leg):
+    """
+    Extract mode, route and attributes as available.
+
+    There are four known cases:
+
+    === Unrouted ===
+
+    For example a leg missing both attributes and route elements, this is the case for non experiences or route plans:
```
alex-kaye, line 336: *non experienced

pam/read/matsim.py
```diff
@@ -0,0 +1,520 @@
...
+        - trav_time
+        - distance
+    The network route is empty.
+
+    Args:
+        leg (xml_leg_element)
+
+    Returns:
+        mode (str), route (pam.activity.Route), attributes (dict)
+    """
```
alex-kaye, line 402: This explanation is v useful!

alex-kaye (approved):
Just some minor grammatical changes. The overall structure looks good to me

pam/read/matsim.py
fredshone: I actually do this by purpose for multi-line comments (ie those between `"""`) - otherwise they are fomratted wrong in stdout

pam/cli.py
```diff
@@ -106,15 +146,24 @@ def summary(
     logger.debug(f"Sample size = {sample_size}.")
     logger.debug(f"MATSim version set to {matsim_version}.")
     logger.debug(f"'household_key' set to {household_key}.")
+    logger.debug(f"Simplify PT trips = {simplify_pt_trips}")
+    print(f"Simplif PT trips = {simplify_pt_trips}")
```
panostsolerid, line 150: Simplify (y at the end)

pam/cli.py
```diff
@@ -190,20 +239,42 @@ def benchmarks(
...
+    "--crop/--no_crop",
+    default=True,
+    help="Crop or don't crop plans to 24 hours, defaults to crop."
+    )
 def stringify(
     path_population_input: str,
     colour: bool,
     width: int,
+    simplify_pt_trips: bool,
+    crop: bool,
```
panostsolerid, line 258: Remove "," at the end

pam/cli.py
```diff
@@ -262,7 +335,12 @@ def crop(
         population = read.read_matsim(
             path_population_input,
             household_key=household_key,
-            version=matsim_version
+            version=matsim_version,
+            simplify_pt_trips=simplify_pt_trips,
+            autocomplete=autocomplete,
+            crop=crop,
+            leg_attributes=leg_attributes,
+            leg_route=leg_route,
```
panostsolerid, line 343: remove "," at the end

pam/cli.py
```diff
@@ -327,7 +415,13 @@ def combine(
...
         combined_population = pop_combine(
             inpaths=population_paths,
-            matsim_version=matsim_version
+            matsim_version=matsim_version,
+            household_key=household_key,
+            simplify_pt_trips=simplify_pt_trips,
+            autocomplete=autocomplete,
+            crop=crop,
+            leg_attributes=leg_attributes,
+            leg_route=leg_route,
```
panostsolerid, line 424: remove ","

pam/cli.py
```diff
@@ -390,14 +487,24 @@ def sample(
...
             path_population_input,
             household_key=household_key,
             weight=1,
-            version=matsim_version
+            version=matsim_version,
+            simplify_pt_trips=simplify_pt_trips,
+            autocomplete=autocomplete,
+            crop=crop,
+            leg_attributes=leg_attributes,
+            leg_route=leg_route,
```
panostsolerid, line 507: remove "," at the end. Sorry for being pedantic!

pam/cli.py
```diff
@@ -106,15 +146,24 @@ def summary(
...
         population = read.read_matsim(
             path_population_input,
             household_key=household_key,
             weight=int(1/sample_size),
             version=matsim_version,
+            simplify_pt_trips=simplify_pt_trips,
+            autocomplete=autocomplete,
             crop=crop,
+            leg_attributes=leg_attributes,
+            leg_route=leg_route,
```
panostsolerid, line 166: remove "," at the end

pam/cli.py
```diff
@@ -67,30 +115,22 @@ def report():
...
     matsim_version: int,
     attribute_key: str,
     household_key: str,
+    simplify_pt_trips: bool,
+    autocomplete : bool,
     crop: bool,
+    leg_attributes: bool,
+    leg_route: bool,
     debug: bool,
     rich: bool,
```
panostsolerid, line 135: remove "," (?)

pam/write/diary.py
```diff
@@ -0,0 +1,204 @@
...
+    hhs = []
+    people = []
+    acts = []
+    legs = []
+
+    for hid, hh in population.households.items():
+        hh_data = {
+            'hid': hid,
+            'freq': hh.freq,
+            'hzone': hh.location.area,
```
panostsolerid, line 41: Ok i've read about the "trailing comma" thing. Maybe it is a good idea after all! for adding lines in the future etc. So ignore my "comma" comments please.

pam/write/diary.py
```diff
@@ -0,0 +1,204 @@
...
+    if 'geometry' in df.columns:
+        df = gp.GeoDataFrame(df, geometry='geometry')
+        if crs is not None:
+            df.crs = crs
+            df.to_crs(to_crs, inplace=True)
+        df.to_file(path, driver='GeoJSON')
+
+
+def save_csv(df, path):
+    """Write GeoDataFrame as csv by droppoing geometry column"""
```
panostsolerid, line 159: dropping

pam/read/diary.py
```diff
@@ -82,10 +77,10 @@ def load_travel_diary(
             logger.warning("Using tour based purpose parser (recommended)")
         else:
             logger.warning(
-                """
-                Using simple trip based purpose parser using, this assumes first activity is 'home'.
-                If you do not wish to assume this, try setting 'tour_based' = True (default).
-                """
+"""
+Using simple trip based purpose parser using, this assumes first activity is 'home'.
```
panostsolerid, line 81: remove second "using"

pam/read/matsim.py
```diff
@@ -0,0 +1,520 @@
...
+        autocomplete : bool = True,
+        crop : bool = False,
+        keep_non_selected : bool = False,
+        leg_attributes : bool = True,
+        leg_route : bool = True,
+):
+    """
+    Load a MATSim format population into core population format.
+    It is possible to maintain the unity of housholds using a household uid in
+    the attributes input, ie:
```
panostsolerid, line 32: i.e.

pam/read/matsim.py
```diff
@@ -0,0 +1,520 @@
...
+    This is a transit routed leg with the route encoded as json string and routingMode attribute:
+        <leg mode="pt" trav_time="00:43:42">
+            <attributes>
+                <attribute name="routingMode" class="java.lang.String">bus</attribute>
+            </attributes>
+            <route type="default_pt" start_link="1-2" end_link="3-4" trav_time="00:43:42" distance="10100.0">
+            {"transitRouteId":"work_bound","boardingTime":"07:30:00","transitLineId":"city_line","accessFacilityId":"home_stop_out","egressFacilityId":"work_stop_in"}
+            </route>
+        </leg>
+    Route must be transit ie there will not be a network route.
```
panostsolerid, line 351: i.e.

panostsolerid (approved):
Just some minor grammatical comments, nothing important at all. I started commenting on trailing commas, but later found about their purpose and I kinda like them now! So please ignore any comments on them.

pam/write/diary.py
fredshone: yeah i like leaving a trailing comma

PR #168, opened 2022-09-13
Slimmer notebooks

fredshone:
for merging after #167 

I've done a little spring cleaning of notebooks. trying to reduce run times and general usability.

Running (on my machine) in about 6 minutes...
image

There's still scope for improvements to content and perhaps some of these can be removed.


Theodore-Chatziioannou (changes requested):
A couple of requests:
- could you drop the changes that are not part of this PR? Of the 62 files changed, I believe you only meant to change the files under the "examples" folder? 
- before we merge, can we make sure that the saved state of the notebooks shows all cell outputs correctly? Ie avoid broken views like [this one](https://github.com/arup-group/pam/blob/46034380214ca1d1953db9f5c847a0456b06afb1/examples/02_Population_Basics.ipynb).

fredshone:
@Theodore-Chatziioannou I've run the notebooks - they look ok now (I added a script).

I've rebased to main so there are less changes in the diff. But there are some changes associated with the notebooks, mainly two types:
- input example data - to reduce notebook run times
- minor code changes to clean up some rough edges in the notebooks (mostly for validation of plans - I slightly relaxed the requirements)

There might also be a few other bits and bobs snuck in of course. But I don't think the code changes really require review. Mainly want you guys to have sight of the notebook cull.

PR #170, opened 2022-10-12
Make the notebooks smoke test faster

mfitz:
As per [LAB-1905](https://arupdigital.atlassian.net/browse/LAB-1905); the smoke test is more than 3 times faster now when running locally. The speed improvement when running inside the CI build is smaller (more like 2 X), probably because there are fewer CPU cores available inside the container running the GitHub Action.

### Before this change

Running the smoke test locally took around 150 seconds
Screen Shot 2022-10-12 at 15 54 32

### After this change

Running the smoke test locally takes around 45 seconds
 
 Screen Shot 2022-10-12 at 15 46 03
 
 ### Verifying failure functionality
 
 Just to make sure that failed processes are still detected, and that the parent process still produces the correct shell return code when one or more notebooks fail, I deliberately broke one of the notebooks. The failure is still detected, and the process return code is non-zero as it should be (which would break the CI build, as desired).
 
Screen Shot 2022-10-12 at 16 41 46

### DIFF
--- a/scripts/code-qa/notebook_smoke_tests.py
+++ b/scripts/code-qa/notebook_smoke_tests.py
@@ -1,8 +1,11 @@
 import argparse
 import glob
+import multiprocessing
 import subprocess
 import sys
 from datetime import datetime
+from itertools import repeat
+from multiprocessing import Manager
 from pprint import pprint
 
 from colorama import Fore, Style
@@ -65,7 +68,7 @@ def install_ipython_kernel(kernel_name):
     return run_shell_command(kernel_install_cmd)
 
 
-def execute_notebook(notebook_path):
+def execute_notebook(notebook_path, notebook_results_dict):
     print("Executing notebook '{}{}{}'...".format(Fore.YELLOW, notebook_path, Style.RESET_ALL))
     execute_notebook_cmd = [
         'jupyter', 'nbconvert',
@@ -74,7 +77,8 @@ def execute_notebook(notebook_path):
         '--execute', '"{}"'.format(notebook_path),
         '--output-dir=/tmp'
     ]
-    return run_shell_command(execute_notebook_cmd)
+    return_code, cmd, run_time = run_shell_command(execute_notebook_cmd)
+    notebook_results_dict[notebook_path] = (return_code, run_time)
 
 
 def run_shell_command(shell_cmd):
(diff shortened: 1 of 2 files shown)


scripts/code-qa/notebook_smoke_tests.py
```diff
@@ -74,15 +77,22 @@ def execute_notebook(notebook_path):
...
+    notebook_results_dict[notebook_path] = (return_code, run_time)
 
 
 def run_shell_command(shell_cmd):
     print(Fore.BLUE + ' '.join(shell_cmd))
     start_time = datetime.now()
     rc = subprocess.call(' '.join(shell_cmd), shell=True)
     running_time = datetime.now() - start_time
-    print("{}Shell process return value was {}{}{}".format(Style.RESET_ALL, Fore.YELLOW, rc, Style.RESET_ALL))
+    print("{}Shell process return value for {}{}{} was {}{}{}".format(Style.RESET_ALL,
```
mfitz, line 89: Previously the notebooks ran in sequential order, so it was easy to understand which process return code matched which notebook. Now the notebooks are running in parallel, we need to explicitly match the return code to the command that produced it, because messages to stdout are now interleaved from different notebooks.

scripts/code-qa/notebook_smoke_tests.py
```diff
@@ -163,11 +173,8 @@ def install_kernels(kernel_list):
     install_kernels(command_args['kernel_name'])
     print("")
 
-    notebook_results = {}
-    for notebook in notebooks:
-        print('------------------------------------------------------')
-        return_code, cmd, run_time = execute_notebook(notebook)
-        notebook_results[notebook] = (return_code, run_time)
-
+    notebook_results = Manager().dict()
```
mfitz, line 176: If you don't create the dictionary this way, each process gets its own copy that disappears when the process terminates, making it impossible to collect the results for later display. `Manager().dict()` creates a _shared_ dict.

scripts/code-qa/notebook_smoke_tests.py
```diff
@@ -163,11 +173,8 @@ def install_kernels(kernel_list):
...
+    with multiprocessing.Pool() as pool:
```
mfitz, line 177: By default, the pool size is scaled according to the number of CPU cores available to the Python runtime.

fredshone (approved):
Zing

Theodore-Chatziioannou (approved):

PR #183, opened 2023-04-05
fix bug

Theodore-Chatziioannou:
Fixes the mysterious xml parsing bug. 

After [this](https://github.com/arup-group/pam/commit/06e0dbc87954079ee763f2a442a6842378c57516) commit, we were getting strange memory errors when trying to read a population created with the same version. However, the only thing that the commit was changing in the xml structure was some rearranging of the comments structure, which shouldn't be creating any errors. 

However, our element parser had a bug, which was not showing itself as long as we were keeping an implicit assumption that the first nodes of the population xml tree were comments. If they were not, the parser was ending up deleting the same element it was streaming, eventually leading to a memory error.

As documented in the relevant [lxml documentation](https://lxml.de/3.2/parsing.html#iterparse-and-iterwalk) (section "Modifying the tree"), we "must not modify or move the ancestors (parents) of the current element. You should also avoid moving or discarding the element itself. ".

This PR fixes the error, changes some of the test fixture data to the "unhappy-path" version, and adds a test.

### DIFF
--- a/pam/utils.py
+++ b/pam/utils.py
@@ -154,7 +154,8 @@ def parse_elems(target, tag):
     for _, element in doc:
         yield element
         element.clear()
-        del element.getparent()[0]
+        while element.getprevious() is not None:
+            del element.getparent()[0]
     del doc
 
 
--- a/tests/test_00_utils.py
+++ b/tests/test_00_utils.py
@@ -13,6 +13,9 @@ from datetime import datetime
 test_trips_path = os.path.abspath(
     os.path.join(os.path.dirname(__file__), "test_data/test_matsim_plans.xml")
 )
+test_tripsv12_path = os.path.abspath(
+    os.path.join(os.path.dirname(__file__), "test_data/test_matsim_plansv12.xml")
+)
 
 
 @pytest.fixture()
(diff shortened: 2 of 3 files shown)


Theodore-Chatziioannou:
> why use `while` and not `if`? This only sees one element at a time.
> 
> otherwise no complaints - great fix and coverage.

we don't need it right now, but it felt a bit safer for future changes. According to the lxml documenation:
```
The while loop deletes multiple siblings in a row. This is only necessary if you skipped over some of them using the tag keyword argument. Otherwise, a simple if should do.
```

tests/test_00_utils.py
```diff
@@ -221,3 +224,15 @@ def test_matsim_time_day_three():
     """ Day-three matsim timestamp """
     dt = utils.matsim_time_to_datetime('49:01:02')
     assert dt == datetime(1900, 1, 3, 1, 1, 2)
+
+def test_parser_does_not_delete_current_element():
```
mfitz, line 228: Presumably, this test failed before you introduced the fix?

Theodore-Chatziioannou: yes, it did

pam/utils.py
```diff
@@ -154,7 +154,8 @@ def parse_elems(target, tag):
     for _, element in doc:
         yield element
         element.clear()
-        del element.getparent()[0]
+        while element.getprevious() is not None:
+            del element.getparent()[0]
```
mfitz, line 158: Do `element.getparent()[0]` and `element.getprevious()` point to the same node when we reach line 158?

Theodore-Chatziioannou: not necessarily, because we can skip nodes according to the tag filter. This is why having commend nodes was not breaking the parsing process in the past.

mfitz (approved):
Good investigation!

It feels somehow wrong to modify the structure of a thing while iterating over that thing. Lots of data structures in some languages won't even allow you to do that, because it introduces a certain amount of jeopardy.

However, I recognise that this is a documented approach to memory efficiency when parsing large XML files with `lxml`, and we've made it safer by explicitly checking our assumption about what we're deleting, so I can live with it 👍

PR #184, opened 2023-04-12
Plan clustering

Theodore-Chatziioannou:
Adds a plan clustering module for grouping similar activity sequences. An "encoder" module converts PAM plans to character sequences, and an agglomeration clustering model is fitted using the edit distance between those sequences.

Introduces some new library dependencies, which -however- will only be used by few users doing more advanced synthesis. I have therefore kept them as "extras" (ie a "full" version of PAM can be installed with `pip install -e .[full]` if a user wishes to work with the planner module).

### DIFF
--- a/pam/planner/clustering.py
+++ b/pam/planner/clustering.py
@@ -1,11 +1,11 @@
-from Levenshtein import ratio, hamming
+from Levenshtein import ratio
 import numpy as np
 from sklearn.cluster import AgglomerativeClustering
 from typing import List, Optional
-from pam.core import Population, Household, Person
+from pam.core import Population
 from pam.activity import Plan
 import pandas as pd
-from functools import lru_cache, cached_property
+from functools import cached_property
 from pam.planner.encoder import PlansCharacterEncoder
 from pam.plot.plans import plot_activity_breakdown_area, plot_activity_breakdown_area_tiles
 import itertools
@@ -21,7 +21,7 @@ def _levenshtein_distance(a: str, b: str) -> float:
 
 def calc_levenshtein_matrix(x: List[str], y: List[str]) -> np.array:
     """
-    Create a levenshtein distance matrix from two lists of strings
+    Create a levenshtein distance matrix from two lists of strings.
     """
     levenshtein_distance = np.vectorize(_levenshtein_distance)
     distances = levenshtein_distance(np.array(x).reshape(-1, 1), np.array(y))
@@ -51,7 +51,7 @@ class PlanClusters:
     @cached_property
     def distances(self) -> np.array:
         """
-        Levenshtein distances between activity plans
+        Levenshtein distances between activity plans.
         """
         dist = calc_levenshtein_matrix(
             self.plans_encoded, self.plans_encoded)
@@ -69,8 +69,10 @@ class PlanClusters:
             linkage: str = 'complete',
     ) -> None:
         """
+        Fit an agglomerative clustering model.
+
         :param n_clusters: The number of clusters to use.
-        :param linkage: Linkage criterion
+        :param linkage: Linkage criterion.
         """
         model = AgglomerativeClustering(
             n_clusters=n_clusters,
--- a/pam/planner/clustering.py
+++ b/pam/planner/clustering.py
@@ -80,7 +80,7 @@ class PlanClusters:
         model = AgglomerativeClustering(
             n_clusters=n_clusters,
             linkage=linkage,
-            affinity='precomputed' # change argument to "metric" for sklearn version>=1.4
+            affinity='precomputed'  # change argument to "metric" for sklearn version>=1.4
         )
         model.fit((self.distances))
 
@@ -109,7 +109,7 @@ class PlanClusters:
         Get the number of plans in each cluster.
         """
         return pd.Series(self.model.labels_).value_counts()
-    
+
     def get_cluster_membership(self) -> dict:
         """
         Get the cluster membership of each person in the population.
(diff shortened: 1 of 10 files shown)


pam/planner/clustering.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+        dist = calc_levenshtein_matrix(
+            self.plans_encoded, self.plans_encoded)
+        return dist
+
+    @property
+    @lru_cache()
+    def distances_no_diagonal(self) -> np.array:
+        dist = self.distances.copy()
+        np.fill_diagonal(dist, 1)
```
fredshone, line 66: why 1?

Theodore-Chatziioannou: this method creates a distance matrix to get a plan's "nearest neighbours". So, we set the distance to 1 (the maximum possible value) to avoid getting the same plan back.

pam/planner/clustering.py
```diff
@@ -0,0 +1,164 @@
...
+        return [self.plans[x] for x in idx_closest]
+
+    def get_cluster_plans(self, cluster: int):
+        """
+        Get the plans that belong in a specific cluster.
+
+        :param cluster: The cluster index.
+        """
+        return list(
+            itertools.compress(self.plans, self.model.labels_ == cluster)
```
fredshone, line 104: curious if using itertools is faster than simple list comprehension? Because I have to look up docs for itertools.compress which makes this less readable.

setup.py
```diff
@@ -24,5 +24,8 @@
     ],
     python_requires='>=3.7',
     entry_points={"console_scripts": ["pam = pam.cli:cli"]},
+    extras_require = {
+        'full': ['scikit-learn', 'python-Levenshtein']
+    },
```
fredshone, line 29: maybe better to call "clustering" rather than "full"? Just incase someone adds tensorflow or something else heavy in future for something else?

pam/planner/clustering.py
Theodore-Chatziioannou: yes, itertools was more performant. But don't mind changing it if you think it is very obscure. Itertools is part of the standard library through, and provides some elegant tools, I think we could use them more

setup.py
Theodore-Chatziioannou: agreed - I wasn't sure about the name either. Perhaps I can call it "planner", which is more clear about the module's purpose.

pam/planner/clustering.py
Theodore-Chatziioannou: > nice, minor comments and questions. I like that you make sk and levenstein optional additionals - but how is this handled when pip installs the clustering module which has these reqs (I don't see this module being declared optional)?

if you run the clustering methods with "vanilla" PAM, you will get an error. I can check whether I can make the error message more useful , ie "Dependency error. To use the Planner module, please install the full PAM version with `pip install -e .[planner]` . What do you think?

Theodore-Chatziioannou:
> 

if you run the clustering methods with "vanilla" PAM, you will get an error. I can check whether I can make the error message more useful , ie "Dependency error. To use the Planner module, please install the full PAM version with pip install -e .[planner] . What do you think?

fredshone:
> if you run the clustering methods with "vanilla" PAM, you will get an error. I can check whether I can make the error message more useful , ie Dependency error. To use the Planner module, please install the full PAM version with pip install -e .[planner] . What do you think?

I think that's a very nice idea.

pam/planner/clustering.py
```diff
@@ -0,0 +1,164 @@
...
+def calc_levenshtein_matrix(x: List[str], y: List[str]) -> np.array:
+    """
+    Create a levenshtein distance matrix from two lists of strings.
+    """
+    levenshtein_distance = np.vectorize(_levenshtein_distance)
+    distances = levenshtein_distance(np.array(x).reshape(-1, 1), np.array(y))
+    return distances
+
+
+class PlanClusters:
```
syhwawa, line 31: Maybe add a doc string to explain a bit about the `PlanClusters` Class?

pam/planner/encoder.py
```diff
@@ -0,0 +1,193 @@
+from __future__ import annotations
+from typing import TYPE_CHECKING
+if TYPE_CHECKING:
```
syhwawa, line 3: Just curious why add if statement here.

pam/planner/encoder.py
```diff
@@ -0,0 +1,193 @@
...
+            start_time += duration
+
+        return plan
+
+    @staticmethod
+    def get_seq(x):
+        raise NotImplementedError
+
+
+class PlanCharacterEncoder(PlanEncoder):
```
syhwawa, line 118: Try to understand the difference between `PlanCharacterEncoder` and `PlanOneHotEncoder `. 
It seems like the `PlanCharacterEncoder` encodes PAM plans into a character sequence used when the encodes the activity during clustering,   while the `PlanOneHotEncoder ` encodes PAM plans into a 2D array and is used in the plot functions.

syhwawa (commented):
The clustering module looks nice to me!  I have played this module and generated some breakdown activities plots using the plot function and it performs well👍.
Left some very minor comments.

setup.py
Theodore-Chatziioannou: done

pam/planner/clustering.py
Theodore-Chatziioannou: done

pam/planner/encoder.py
Theodore-Chatziioannou: to avoid circular dependencies when doing type checking with classes that have not yet been initialised

pam/planner/encoder.py
Theodore-Chatziioannou: exactly. OneHotEncoder converts the plan to a 2D boolean array. Currently only used by the plot function, but will have more applications later

PR #189, opened 2023-05-23
Choice models

Theodore-Chatziioannou:
Introduces abstract mode and location choice modules.

The pam.planner.choice class allows the user to apply a choice model for selecting the location of activities and the mode for accessing them, given person characteristics, network conditions and/or zone attraction data. The module is designed to be extendable, supporting different kinds of choice models and sampling approaches.

The typical workflow goes as follows:

```
choice_model = ChoiceMNL(population, od, zones) # initialize the model and point to the data objects 
choice_model.configure(u, scope) # configure the model by specifying a utility function and the scope of application.
choice_model.apply() # apply the model and update the population with the results.
```

The module design is shown below:

planner_choice_design

We use the planner.choice.Choice class as a mediator between the population, data, and configuration.
* The `od` object manages origin-destination data, for example travel time and travel distance between each origin and destination, for each travel mode.
* The `zones` dataset includes destination attraction data, for example the number of jobs or schools in each likely destination zone.
* Population is a PAM population (`pam.core.Population`) object

We configure the model by specifying:

- the scope of the model. For example, work activities.
- the utility formulation of each alternative.

Both settings are defined as strings. The strings may comprise mathematical operators, coefficients, planner data objects (od / zones), and/or PAM population objects (person/ act).

Coefficients can be passed either as a number, or as a list, with each element in the list corresponding to one of the modes in the od object.

An example application is presented in this notebook: `examples/14_Advanced_Plan_Cropping.ipynb`.

### DIFF
--- a/pam/planner/od.py
+++ b/pam/planner/od.py
@@ -1,0 +1,74 @@
+"""
+Manages origin-destination data required by the planner module.
+"""
+import numpy as np
+from typing import Union, Optional, List, NamedTuple
+
+
+class Labels(NamedTuple):
+    """ Data labels for the  """
+    mode: List
+    vars: List
+    origin_zones: List
+    destination_zones: List
+
+
+class OD:
+    """
+    Holds origin-destination matrices for a number of modes and variables.
+    """
+    dimensions = ['mode', 'variable', 'origin', 'destination']
+
+    def __init__(
+        self,
+        data: np.ndarray,
+        labels: Union[Labels, List, dict]
+    ) -> None:
+        """
+        :param data: A multi-dimensional numpy array of the origin-destination data.
+            - First dimension: mode (ie car, bus, etc)
+            - Second dimension: variable (ie travel time, distance, etc)
+            - Third dimension: origin zone
+            - Fourth dimension: destination zone
+        """
+        self.data = data
+        self.labels = self.parse_labels(labels)
+        self.data_checks()
+
+    def data_checks(self):
+        """
+        Check the integrity of input data and labels.
+        """
+        assert self.data.ndim == 4, \
+            "The number of matrix dimensions should be 4 (mode, variable, origin, destination)"
+        for i, (key, labels) in enumerate(zip(self.labels._fields, self.labels)):
+            assert len(labels) == self.data.shape[i], \
+                f"The number of {key} labels should match the first dimension of the OD dataset"
+            
+    @staticmethod
+    def parse_labels(labels: Union[Labels, List, dict]) -> Labels:
+        """
+        Parse labels as a named tuple.
+        """
+        if not isinstance(labels, Labels):
+            if isinstance(labels, list):
+                return Labels(*labels)
+            elif isinstance(labels, dict):
+                return Labels(**labels)
+            else:
+                raise ValueError('Please provide a valid label type')
+        return labels
+    
+    def __getitem__(self, args):
+        _args = args if isinstance(args, tuple) else tuple([args])
+        _args_encoded = tuple()
+        for i, (arg, labels) in enumerate(zip(_args, self.labels)):
+            if arg == slice(None) or isinstance(arg, int):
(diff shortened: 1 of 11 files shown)


pam/planner/od.py
```diff
@@ -0,0 +1,84 @@
...
+        self,
+        data: np.ndarray,
+        labels: Union[Labels, List, dict]
+    ) -> None:
+        """
+        :param data: A multi-dimensional numpy array of the origin-destination data.
+            - First dimension: mode (ie car, bus, etc)
+            - Second dimension: variable (ie travel time, distance, etc)
+            - Third dimension: origin zone
+            - Fourth dimension: destination zone
```
fredshone, line 32: bit scary assuming the order of the dimension in the ndarray. I already got confused by the example which declares the labels:
```
labels = {
    'mode': ['car', 'bus'],
    'vars': ['time', 'distance'],
    'origin_zones': ['a', 'b'],
    'destination_zones': ['a', 'b']
}
```

In a different order to the structure (`variables, origin zone, destination zone, and mode`) and a different order to the str representation:

```
Origin-destination dataset 
--------------------------------------------------
Labels(vars=['time', 'distance'], origin_zones=['a', 'b'], destination_zones=['a', 'b'], mode=['car', 'bus'])
--------------------------------------------------
time - car:
[[20 40]
 [40 20]]
--------------------------------------------------
time - bus:
[[30 45]
 [45 30]]
--------------------------------------------------
distance - car:
[[5 8]
 [8 5]]
--------------------------------------------------
distance - bus:
[[5 9]
 [9 5]]
--------------------------------------------------
```

Is there a more common structure to these arrays? For example i am more use to mode being a higher dimension than o and d.

On any case might be worth being more explicit about short dimensions (such as mode and variable) with named arrays or more simply dictionaries? eg:

```
{
  'time': {
      'a': {
        'a': {'car': 20, 'bus': 30},
        'b': {'car': 40, 'bus': 45}
       },
      'b': {
        'a': ['car': 40, 'bus': 45],
        'b': ['car': 20, 'bus': 30]
       }
     },
   'distance': {
      'a': {
         'a': ['car': 5,  'bus': 5],
         'b': ['car': 8,  'bus': 9]
        },
      'b': {
         'a': ['car': 8,  'bus': 9],
         'b': ['car': 5,  'bus': 5]
        }
    }
}
```

or:

```
{'time': {
  'car': ndarray([[],[],...) // ods,
  'bus': ndarray([[],[],...) // ods
...
```

Theodore-Chatziioannou: my docstring there was wrong. The matrix dimensions are always `(variables, origin zone, destination zone, and mode)`, in that order. 

This order of dimensions (with mode at the end) is indeed a bit unusual, but allows for easier (and more abstract) specification of the choice model. More specifically, it allows us to pass coefficients as lists, with each list corresponding to each mode. For example `u = '[0, -1] * od['time']`, which gets interpreted as `beta_mode0= 0, beta_mode1= -1`.

Holding our data as numpy arrays also allows for good performance, which is important given the size our typical inputs.

Theodore-Chatziioannou: however, I agree with your comment, it is hard to construct that 4D array, and the flow may break easily. I have therefore now introduced a factory method `plan.planner.od.ODFactory.from_matrices(matrices)` , where matrices have this form:

```
matrices = [
    ODMatrix(var='time', mode='car', origin_zones=['a','b'], destination_zones=['a','b'], matrix=np.array([[20, 40], [40, 20]])),
    ....
]
```

The factory also includes a number of data integrity checks and returns resulting the `OD` object.

Theodore-Chatziioannou: this is similar to your last suggestion in the comment, with a namedtuple instead of dictionary to ensure the keys are passed correctly.

fredshone (approved):

panostsolerid (approved):

Issue #144, opened 2022-05-13
Include full linting checks in the code QA script


mfitz:
Currently we are running a code linting check in the CI build that restricts the linting rules in play via the argument `--select=E9,F63,F7,F82`. We should be using the full default set of rules so that we catch and fix a wider range of linting problems. Using this full set of rules will require us to fix a few hundred linting errors:

```bash
flake8 . --max-line-length 120 --count --show-source --statistics --exclude=scripts,tests
...
2     E101 indentation contains mixed spaces and tabs
12    E122 continuation line missing indentation or outdented
3     E124 closing bracket does not match visual indentation
33    E125 continuation line with same indent as next logical line
4     E128 continuation line under-indented for visual indent
61    E203 whitespace before ':'
4     E221 multiple spaces before operator
20    E225 missing whitespace around operator
1     E227 missing whitespace around bitwise or shift operator
1     E228 missing whitespace around modulo operator
95    E231 missing whitespace after ':'
271   E251 unexpected spaces around keyword / parameter equals
21    E261 at least two spaces before inline comment
6     E262 inline comment should start with '# '
9     E265 block comment should start with '# '
13    E266 too many leading '#' for block comment
1     E271 multiple spaces after keyword
22    E302 expected 2 blank lines, found 1
10    E303 too many blank lines (2)
52    E501 line too long (150 > 120 characters)
1     E502 the backslash is redundant between brackets
12    E711 comparison to None should be 'if cond is not None:'
1     E712 comparison to True should be 'if cond is True:' or 'if cond:'
3     E713 test for membership should be 'not in'
22    E999 SyntaxError: invalid syntax
5     F401 'pam.plot.plans.*' imported but unused
5     F403 'from pam.plot.plans import *' used; unable to detect undefined names
2     W191 indentation contains tabs
36    W291 trailing whitespace
4     W292 no newline at end of file
34    W293 blank line contains whitespace
1     W391 blank line at end of file
767
```

We should extract the linting into its own script, and invoke this script from our [new QA script/git commit hook](https://github.com/arup-group/pam/blob/8ee59a852b982b04bce91d00482608512a28f19e/scripts/code-qa/qa-checks.sh).

PR #192, opened 2023-06-12
Update requirements to make PAM packagable

### LINKED ISSUE #178: pytest fails when using python 3.11 (and updated dependencies)
I tried to use pam with python 3.11 - generating a population (mostly with the facility sampler) worked fine and was roughly 20% faster than using python 3.7 (wohoo!)

-> See https://github.com/markusstraub/pam/tree/py311, especially the `requirements.txt` with all current dependencies

However I noticed that a few unit tests fail. This is (I guess) due to the updated (geo)pandas-dependencies:

```
FAILED tests/test_00_utils.py::test_build_geodataframe_for_pt_person - AssertionError: Attributes of DataFrame.iloc[:, 11] (column name="seq") are different
FAILED tests/test_00_utils.py::test_build_geodataframe_for_cyclist - AssertionError: Attributes of DataFrame.iloc[:, 11] (column name="seq") are different
FAILED tests/test_00_utils.py::test_build_hhld_geodataframe - AssertionError: Attributes of DataFrame.iloc[:, 11] (column name="seq") are different
FAILED tests/test_00_utils.py::test_build_pop_geodataframe - AssertionError: Attributes of DataFrame.iloc[:, 11] (column name="seq") are different
FAILED tests/test_09_samplers.py::test_random_sample_point_from_multilinestring_random_seed - ValueError: Total of weights must be greater than zero
FAILED tests/test_09_samplers.py::test_random_sample_point_from_multipoint_random_seed - ValueError: Total of weights must be greater than zero
```

The fails in `test_00_utils.py` don't worry me at all, apparently pandas 1.3.x converted the `seq` parsed from the matsim xml to float, with pandas 1.5.x it is an int.

The fails in `test_09_samplers.py` are strange, but maybe these two tests need to be fixed anyways:
- `test_random_sample_point_from_multilinestring_random_seed` calls `sampler.sample_point_from_multipolygon`
- `test_random_sample_point_from_multipoint_random_seed` calls `sampler.sample_point_from_multipolygon`

@YannisZa git tells me you wrote (or last edited) these two tests, do you have an idea what could be wrong here?

brynpickering:
I have split up the requirements to ensure only those needed by the package can be installed as a basis. The others are now split into "docs" (for the notebooks) and "tests".

I've also added a conda environment file, although it is mostly duplication of the requirements text files and can be removed once s2sphere is available on the conda-forge channel (https://github.com/conda-forge/staged-recipes/pull/23041). 

Incidently fixed #178 to ensure tests pass. There were copy-paste errors in some of the tests.


tests/test_00_utils.py
```diff
@@ -147,22 +150,25 @@ def correct_cyclist_geodataframe():
...
-    assert [[(round(p[0], 6), round(p[1], 6)) for p in ls.coords] for ls in gdf['geometry'].to_list()] == [
+    expected_coords = [
         [(0.123336, 51.537168), (0.175434, 51.567355)], [(0.175434, 51.567355), (0.183252, 51.574827)],
         [(0.183252, 51.574827), (0.250898, 51.559106)], [(0.250898, 51.559106), (0.250894, 51.559016)],
         [(0.250894, 51.559016), (0.080917, 51.539493)], [(0.080917, 51.539493), (0.080917, 51.539493)],
-        [(0.080917, 51.539493), (0.146117, 51.526088)], [(0.146117, 51.526088), (0.123336, 51.537168)]]
+        [(0.080917, 51.539493), (0.146117, 51.526088)], [(0.146117, 51.526088), (0.123336, 51.537168)]
+    ]
+    built_coords = [[(round(p[0], 6), round(p[1], 6)) for p in ls.coords] for ls in gdf['geometry'].to_list()]
+    np.testing.assert_allclose(built_coords, expected_coords, rtol=1e-3)
```
brynpickering, line 165: This is a reasonably lax tolerance to make up for the fact that the numbers all changed (potentially) non-negigibly. I'm not familiar enough with the data to know if the numbers have changed enough to suggest a problem. I think it's more likely just a change in the CRS conversion.

tests/test_00_utils.py
```diff
@@ -147,22 +150,25 @@ def correct_cyclist_geodataframe():
...
 
 
 def test_build_geodataframe_for_cyclist(cyclist, correct_cyclist_geodataframe):
     gdf = cyclist.build_travel_geodataframe()
     gdf = gdf[correct_cyclist_geodataframe.columns]
-    assert_frame_equal(gdf, correct_cyclist_geodataframe)
+    assert_frame_equal(gdf, correct_cyclist_geodataframe, check_dtype=False)
```
brynpickering, line 171: This and others below allow a column to be either float and int and still be compared as being equal provided they are the same value (1 == 1.0). Other option would be to change the test geodataframe to have integer "seq" values, but this approach seems more maintainable.

tests/test_09_samplers.py
```diff
@@ -1,10 +1,11 @@
-import pytest
 import random
-import pandas as pd
-import geopandas as gp
-from shapely.geometry import Polygon, Point, LinearRing, LineString, MultiLineString, MultiPolygon, MultiPoint
 from types import GeneratorType
 
+import geopandas as gp
```
brynpickering, line 4: No change in packages, just [sorted](https://pycqa.github.io/isort/)

tests/test_09_samplers.py
```diff
@@ -220,7 +221,7 @@ def test_random_sample_point_from_multilinestring_random_seed(fixed_seed):
     poly = MultiLineString([p1, p2])
     gdf = gp.GeoDataFrame(df, geometry=[poly]*3)
     sampler = spatial.RandomPointSampler(gdf.geometry, patience=0, seed=fixed_seed)
-    assert list(sampler.sample_point_from_multipolygon(gdf.geometry[0]).coords[0]) == [10.134364244112401, 10.847433736937234]
+    assert list(sampler.sample_point_from_multilinestring(gdf.geometry[0]).coords[0]) == [0.5374569764496049, 0.0]
```
brynpickering, line 224: copy-paste error in the method called. Didn't raise an error previously as non-zero areas weren't _required_. An update in geopandas(?) changed that, at which point providing a `multilinestring` to a method expecting a `multipolygon` failed. After fixing the method, I just pasted in the values that comes as a result. Checking the multilinestring, this point is on it.

tests/test_09_samplers.py
```diff
@@ -239,7 +240,7 @@ def test_random_sample_point_from_multipoint_random_seed(fixed_seed):
     poly = MultiPoint([p1, p2])
     gdf = gp.GeoDataFrame(df, geometry=[poly]*3)
     sampler = spatial.RandomPointSampler(gdf.geometry, patience=0, seed=fixed_seed)
-    assert list(sampler.sample_point_from_multipolygon(gdf.geometry[0]).coords[0]) == [10.0, 10.0]
+    assert list(sampler.sample_point_from_multipoint(gdf.geometry[0]).coords[0]) == [0.0, 0.0]
```
brynpickering, line 243: After fixing the method, I just pasted in the values that comes as a result. Checking the multilinestring, this point is on it.

.github/workflows/build_pipeline.yml
```diff
@@ -7,31 +7,17 @@ jobs:
...
-        export C_INCLUDE_PATH=/usr/include/gdal
-        sudo apt-get install ca-certificates
-        export CURL_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt
-        pip install GDAL==3.4.1
-        pip install -e .[planner]
+        micromamba-version: latest
+        environment-file: environment.yml
+        init-shell: bash
+        post-cleanup: all
+        cache-environment: true
```
brynpickering, line 16: I'll admit, I don't know if this cache will pass between the two CI pipelines.

mfitz: We have configured the repo such that CI builds need to be approved for first-time committers, so I guess we'll find out when I approve the build on this branch...

.github/workflows/build_pipeline.yml
```diff
@@ -2,36 +2,28 @@ name: CI
...
+        create-args: flake8
+        post-cleanup: all
+        cache-environment: true
+      
+    - name: install PAM
+      run: pip install .
+
     - name: Lint with flake8
       run: |
-        pip install flake8
```
mfitz, line 34: That reminds me about #144 ...

tests/test_00_utils.py
mfitz: Is this the smallest tolerance that fixes the test?

I agree that this is likely a change in the CRS conversion across library versions. I'm not mad keen on the assertion in the first place, but that gripe predates this PR.

tests/test_00_utils.py
mfitz: I definitely prefer this approach to modifying the dataframe types.

Dockerfile
```diff
@@ -1,14 +1,9 @@
-FROM python:3.7.16-bullseye
+FROM mambaorg/micromamba:1.4.3
```
mfitz, line 1: Glad to see the Dockerfile simplified. Hopefully, the resultant Docker image is smaller too - do you have the before and after stats?

Dockerfile
```diff
@@ -1,14 +1,9 @@
...
-&& /usr/local/bin/python -m pip install --upgrade pip
+RUN pip install -e .
 
-COPY . .
-
-RUN pip install -e .[planner]
-ENV PYTHONPATH=./scripts:${PYTHONPATH}
-
-ENTRYPOINT ["python3"]
+ENTRYPOINT ["/usr/local/bin/_entrypoint.sh", "ipython"]
```
mfitz, line 9: There are no instructions on using the Docker image in the README, and I don't get the impression people are using Dockerised PAM much, so this is not a question of great import, but...

Will this change how a user invokes anything via a container?

mfitz (approved):
Looks good.

We'll see how the CI builds go once it's approved and then merged...

tests/test_00_utils.py
brynpickering: yeah, agreed, it's quite unmaintainable...
I think possibly `4e-4` or something like that is *the* smallest, but I like it being `1e...`

Dockerfile
brynpickering: Yup, it does make it smaller! locally, before=2.31GB, after=1.53GB

Dockerfile
brynpickering: Good question. I didn't have any examples to play with due to the lack of instructions in the README. I'll check with @Theodore-Chatziioannou. 

Anyway, the first element is just a requirement to have the mamba env variables persist (see [here](https://github.com/mamba-org/micromamba-docker#activating-a-conda-environment-for-entrypoint-commands)). I updated to `ipython` as it *should* provide the same and _more_ functionality as `python`.

brynpickering:
I should note @mfitz that the CI builds on my fork are failing (e.g., [this one](https://github.com/brynpickering/pam/actions/runs/5256476241/jobs/9497857105)) due to issues with secrets and AWS access credentials (I assume both would be dealt with in the merge CI run as the secrets are set in this repo) AND coverage.

Locally, I have coverage >89%, but I think the `__init__.py` file I added isn't being correctly ignored by the run in the GH action. I have a PR that would clean up QA and for which the [tests _do_ pass](https://github.com/brynpickering/pam/actions/runs/5256611971/jobs/9498183085)

Theodore-Chatziioannou (approved):
managed to install PAM on Windows and WSL. My only deviation from the readme was using the "miniforge" prompt on Windows, and calling `miniconda` on WSL.

PR #193, opened 2023-06-14
Clean QA 

brynpickering:
I've stripped down the QA scripts to rely on standard python project config files, rather than custom bash scripts, to produce the same result.

Things I've changed:

1. Moved pytest and coverage config to a `pyproject.toml` file, which can be extended in future with other config items (this is now the recommended config filetype for python projects). 
2. Remove notebook smoke testing in favour of the pytest [nbmake plugin](https://github.com/treebeardtech/nbmake). It does the same thing, but is now integrated into pytest automatically. It could also be used to re-run and save the notebooks, but I have it just test that they run successfully.
3. Replaced flake8 with [Ruff](https://beta.ruff.rs/docs/), in part because it can be configured in the pyproject.toml, but also because it is faster and more configurable. It's written in Rust @fredshone ;)
4. Created a pre-commit config file which will run the Black code formatter, lint with Ruff (only matching the subset of issues covered by flake8 in the old GH actions, will extend to all possible linting errors as per #144 in a later PR), and will fail on the existence of large staged files.
5. EDIT: Removed the legacy memory profiling scripts in favour of using [memray](https://github.com/bloomberg/pytest-memray) on an integration test (same script, just moved into the test suite with a `high_mem` marker). This works fine on linux/mac devices, but annoyingly memray isn't available on Windows. @Theodore-Chatziioannou could you check if it works on WSL?

~The only script that I haven't removed (yet) is the memory profiler. One approach would be to add memory limits to certain tests (maybe with some new tests that do what the profiling script currently does) and then have those tests fail using [memray](https://github.com/bloomberg/pytest-memray) if memory consumption goes too high.~

A [pre-commit CI bot](https://pre-commit.ci/) can be added to this repo if we want to check the pre-commit hooks in the case that a user forgets to install it on their device.

Another benefit of the pyproject toml is that it plays nicely with IDE linting/formatting extensions. They will automatically (on VSCode, at least) apply the project config when working in that directory.


tests/test_00_utils.py
brynpickering, line 1: Black formatted the document, hence the large number of line changes. Only changes made were to update some methods to deal with deprecation warnings popping up during tests.

tests/test_09_samplers_tour.py
brynpickering, line 1: Black formatted the document, hence the large number of line changes. Only changes made were to update some methods to deal with deprecation warnings popping up during tests.

tests/test_20_planner_clustering.py
brynpickering, line 1: Black formatted the document, hence the large number of line changes. Only changes made were to update some methods to deal with deprecation warnings popping up during tests.

pam/core.py
brynpickering, line 1: Black formatted the document, hence the large number of line changes. Only changes made were to update some methods to deal with deprecation warnings popping up during tests.

pam/plot/plans.py
brynpickering, line 1: Black formatted the document, hence the large number of line changes. Only changes made were to update some methods to deal with deprecation warnings popping up during tests.

pam/plot/stats.py
brynpickering, line 1: Black formatted the document, hence the large number of line changes. Only changes made were to update some methods to deal with deprecation warnings popping up during tests.

fredshone:
LGTM :)

requirements/dev.txt
```diff
@@ -0,0 +1,7 @@
+pytest ~= 7.3
+pytest-cov ~= 4.1
+pytest-mock ~= 3.10
+pytest-xdist ~= 3.3
+pytest-timeout ~= 2.1
+nbmake ~= 1.3
```
mfitz, line 6: Can we order this file alphabetically?

mfitz (approved):
I was expecting GitHub Actions to run against your branch now that you've committed to `main`. Maybe it didn't because you had not yet committed to `main` at the time you opened the PR?

It's quite difficult to know what will happen with the Actions until we see them running, but I think we're okay to merge and then fix them up if need be afterwards, just as long as we don't leave the builds in a broken state for very long.

PR #196, opened 2023-06-22
Update example notebooks

brynpickering:
I've cleaned up the notebooks a bit by adding in headers where missing (based on my understanding of the content of the notebooks...) and linting/reformatting the code.

I've also removed all example data that isn't directly used in a notebook. They might be nice to have for when someone *wants* to play with that data, but this isn't really the purpose of a python package repository...

It would be great to also get rid of the LSOA boundary data, because it has a relatively large footprint on disk (~4MB) and is binary data (plain text is better for version control, but an equivalent geojson file has approx. double the filesize). If you don't mind having the example notebook only work while online, we could remove this data and replace it with direct access in the 2 notebooks where it is needed:

``` python
lsoas = gp.read_file("https://data.london.gov.uk/download/statistical-gis-boundary-files-london/9ba8c833-6370-4b11-abdc-314aa020d5e0/statistical-gis-boundaries-london.zip!statistical-gis-boundaries-london/ESRI/LSOA_2004_London_Low_Resolution.shp")
```

- [x] Notebook tests pass


fredshone (approved):
Thank you for looking through these. I skim and have no complaints.

re the lsoas data. I prefer the data to be part of the project. It will be easier to maintain and easier for users to access the examples. 

I've raised issue #200 for Yuhao to reduce the file size. So you can merge this.

PR #197, opened 2023-06-22
Feature: documentation pages

brynpickering:
Here's my first pass at moving the documentation out of the README and into its separate docs page.

You can preview the docs either by following the install isntructions in the README and then running `mike serve`, or by checking out https://brynpickering.github.io/pam . The docs would ultimately be deployed on https://arup-group.github.io/pam, with the HTML stored in a PAM branch called `gh-pages`.

The API reference page shows how poorly docstringed the source code is, but where google-style docstrings have been used (I've updated `pam.core` to this style), it renders reasonably well.


fredshone (approved):
love it. Happy to force this through then start a rewrite

mfitz (approved):

PR #198, opened 2023-06-26
Clean up gitignore

brynpickering:
Removed reference to personal files in the gitignore file and grouped things to ignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,10 @@ __pycache__
 build/
 *.egg-info
 
+# Virtual environment
+venv/
+.venv
+
 # Reports
 .coverage
 .coverage.*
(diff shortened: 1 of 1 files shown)


fredshone (approved):
looks good on my machine although I have a freshish clone.

presumably someone is going to accidentallly add an excel doc in the next few weeks tho...

similarly with venv - people will have stuff hanging around.

Happy to merge but you're gonna need to follow up with some dissemination, maybe easier to just add a group called "legacy"?

brynpickering:
Most of these things should be part of people's global `.gitignore` file, rather than being in the project-specific one (`git config --global core.excludesFile '~/.gitignore'`). 

Otherwise, random notebooks/data files can be caught by a user by not staging them before committing / catching them in PRs.

PR #201, opened 2023-06-27
Reformat code with black and ruff

### LINKED ISSUE #144: Include full linting checks in the code QA script

brynpickering:
Fixes #144.

I have reformatted the code using [black](https://black.readthedocs.io/en/stable/) and [ruff](https://beta.ruff.rs/docs/) and extended the linting rules which will apply in future QA. This brings the code more in line with the [pep8 python code style guidelines](https://peps.python.org/pep-0008/). 

There are *many* more [linting/code style rules](https://beta.ruff.rs/docs/rules/) that could be applied, beyond the base ones I've applied. However, the more that are added, the more that the current code fails to adhere to, requiring manual code edits to fix (rather than what have mostly been automatic fixes on my application of `ruff` here). 

I would say we can extend the applied rules as and when we find time to update the code to adhere to those rules. One obvious one is adding docstrings for all public functions/modules, which we currently don't enforce.

I would also prefer to bring the line length down from 120 to ~90 lines (the pep8 standard is 79, black defaults to 88). Is there a reason for the 120 lines?


pam/cli.py
```diff
@@ -311,15 +245,13 @@ def crop(
     keep_non_selected: bool,
     comment,
     buffer,
-    debug
+    debug,
```
KasiaKoz, line 248: should that comma be there?

pam/samplers/tour.py
```diff
@@ -1,223 +1,208 @@
...
+        hour,
+        minute,
+        o_zone,
+        d_dist,
+        d_freq,
+        facility_sampler,
+        activity_params,
+        threshold_matrix=None,
+        threshold_value=None,
+    ):
```
KasiaKoz, line 204: everywhere else the signatures are collapsed to be inline but it's the opposite here - a bit weird

tests/test_19_combining.py
```diff
@@ -38,4 +38,4 @@ def test_combined_length(path_population_A, path_population_B):
 #         path_boundary,
 #         path_output_dir
 #     )
-#     assert os.path.exists(os.path.join(path_output_dir, 'plans.xml'))
\ No newline at end of file
+#     assert os.path.exists(os.path.join(path_output_dir, 'plans.xml'))
```
KasiaKoz, line 41: surprised this commented out test wasn't deleted

KasiaKoz (approved):
some minor comments

pam/samplers/tour.py
brynpickering: if > 120 lines, it has to break the sigs. No other function in the file has enough sigs to go over 120 lines.

tests/test_19_combining.py
brynpickering: This would be based on running ruff with rule [ERA](https://beta.ruff.rs/docs/rules/#eradicate-era), but it isn't very clever and doesn't properly eradicate the whole commented-out code block. I'll delete it manually.

mfitz:
> I would also prefer to bring the line length down from 120 to ~90 lines (the pep8 standard is 79, black defaults to 88). Is there a reason for the 120 lines?

So I chose 120 characters based on:

- it being the default in PyCharm
- personal preference

My personal preference is based on a line width that uses available screen real-estate, but without having to horizontally scroll, even with two windows of code side-by-side on the screen.

I spent years working on projects that used 80 characters, and over time found that:

- with modern editors and consoles, a lot of horizontal space on the screen goes to waste
- devs sometimes introduce awkward line breaks to fit inside the limit
- devs sometimes choose unhelpful but short variable and function names purely in order to save horizontal space
- attempts to avoid horizontal scrolling lead to more vertical scrolling (ideally, I want to minimise both, which is my aim with 120 characters)

Pep8 does suggest 79 characters, but it also says:

> Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the line length limit up to 99 characters, provided that comments and docstrings are still wrapped at 72 characters.

I think line length is somewhat arbitrary, as the various differing defaults demonstrate. The 80-character limit probably dates back to the optimal width of punchcards, later carried over to terminals. Even I wasn't around for that era...

I recognise that it's sometimes contentious, and personal preferences play a large part. If there is a strong universal desire to move to 80/other characters, I wouldn't object, but if not, I'd rather keep to 120.

brynpickering:
Agreed, mostly personal preference. Avoiding long names and weird line breaking should be solved by just letting automatic formatters deal with all that automatically on commit/save. 

This is why I like 88 characters (the Black default - rulers for 80 and 88 lines shown):

image

On my 14 inch device I can have two files side by side (e.g., source + tests) on one screen. Even without the directory tree on the left, 120 lines would not be reached.

Admittedly, on a 24-inch screen, the 120 line width is fine with two files side-by-side:

image

brynpickering:
Here is the argument for 88 characters: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length

brynpickering:
With my side-by-side on a 14-inch screen argument, I could go up to 100 characters... Maybe a compromise...?

mfitz:
> With my side-by-side on a 14-inch screen argument, I could go up to 100 characters... Maybe a compromise...?

I'm happy with that if nobody else objects 👍

mfitz (approved):

PR #202, opened 2023-06-27
Add codecov to github actions

brynpickering:
This should lead to us having a bot that tells us about coverage changes in the PR, including which files have seen an increase/decrease in coverage.

### DIFF
--- a/.github/workflows/build_pipeline.yml
+++ b/.github/workflows/build_pipeline.yml
@@ -47,6 +47,12 @@ jobs:
 
     - name: Run tests
       run: pytest
+      
+    - name: Upload coverage reports to Codecov
+      uses: codecov/codecov-action@v3
+      env: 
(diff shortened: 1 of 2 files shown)


brynpickering:
The pre-commit CI checks whether all files in the repo are Black and Ruff compliant and that no files are greater than 1MB. It will fail on PRs until #201 is merged in.

mfitz (approved):
Looks good. We experimented with something like this on one of our projects, but never got around to rolling it out widely. It's definitely useful information when reviewing a PR.

The auth cred that is stored on the repo in `secrets.CODECOV_TOKEN` - can we add that to Secrets Manager in our tcm-dev AWS account so that team members have access to it?

brynpickering:
Actually, shouldn't even need the token as it is a public repo. Have now removed it!

PR #205, opened 2023-06-29
Fix coverage reporting

brynpickering:
I realised the config wasn't capturing coverage correctly (and, quite possibly, the old config didn't either). 

In updating the config to *only* focus on the source code + a plugin I wrote for the docs, coverage dropped to 88.something%, under the 89% threshold. I've now added some basic tests for some parts of the `pam/optimise` scripts, since these had _no_ tests so far. This brings it back over the threshold, but we should probably have a target of no single file below 89% coverage, which the repo wouldn't pass right now.

### DIFF
--- a/tests/test_27_optimise_random.py
+++ b/tests/test_27_optimise_random.py
@@ -10,27 +10,40 @@ def stopper():
 
 
 class TestStopper:
-    def test_stopper_init(self, stopper):
+    def test_init(self, stopper):
         assert stopper.record == []
         assert stopper.horizon == 3
         assert stopper.sensitivity == 0.03
 
-    def test_stopper_ok_1(self, stopper):
+    def test_adds_to_record(self, stopper):
         assert stopper.ok(1) is True
         assert stopper.record == [1]
 
-    def test_stopper_ok_2(self, stopper):
-        assert stopper.ok(1) is True
-        assert stopper.ok(1.02) is True
-        assert stopper.record == [1, 1.02]
+    def test_adds_to_already_populated_record(self, stopper):
+        stopper.ok(1)
+        stopper.ok(1.02)
+        stopper.ok(2)
+        assert stopper.record == [1, 1.02, 2]
 
-    def test_stopper_ok_3_no_stop(self, stopper):
+    def test_pops_oldest_value_when_record_exceeds_horizon_length(self, stopper):
         for i in [1, 2, 3, 4]:
-            assert stopper.ok(i) is True
+            stopper.ok(i)
         assert stopper.record == [2, 3, 4]
 
-    def test_stopper_ok_3_stop(self, stopper):
+    def test_ok_is_true_when_sensitivity_value_not_breached(self, stopper):
+        for i in [1, 2, 3, 4]:
+            assert stopper.ok(i) is True
+
+    def test_ok_is_false_when_sensitivity_value_breached_and_record_exceeds_horizon_length(
+        self, stopper
+    ):
         for i in [1, 2, 2.01]:
             assert stopper.ok(i) is True
         assert stopper.ok(2.02) is False
+
+    def test_pops_oldest_when_sensitivity_value_breached_and_record_exceeds_horizon_length(
+        self, stopper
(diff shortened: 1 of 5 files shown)


tests/test_27_optimise_random.py
```diff
@@ -0,0 +1,36 @@
...
+
+from pam.optimise import random
+
+
+@pytest.fixture
+def stopper():
+    return random.Stopper(horizon=3, sensitivity=0.03)
+
+
+class TestStopper:
```
mfitz, line 12: Is there a reason you prefer to wrap test functions in a class?

tests/test_27_optimise_random.py
```diff
@@ -0,0 +1,36 @@
...
+        assert stopper.ok(1) is True
+        assert stopper.ok(1.02) is True
+        assert stopper.record == [1, 1.02]
+
+    def test_stopper_ok_3_no_stop(self, stopper):
+        for i in [1, 2, 3, 4]:
+            assert stopper.ok(i) is True
+        assert stopper.record == [2, 3, 4]
+
+    def test_stopper_ok_3_stop(self, stopper):
```
mfitz, line 32: The test name could do more to describe the behaviour we're verifying here. It's clear to me that we're testing a case when the stopper should call a halt, but it's not clear _why_ we are expecting that (I had to go and read the code under test to understand it).

Could we use the test name to describe the conditions we're testing that cause us to expect stopping?

I'm thinking of something like `test_stops_when_sensitivity_value_breached`, or maybe even a concise name that describes what leads to a breach of the sensitivity value?

As a side note, I'm also wondering if the stopper should explicitly order the values it holds in the list. We're subtracting the first value from the last - does this imply we expect them to be ordered? Does that ordering happen implicitly due to the way the stopper is being used? I'm aware you didn't write that code, but thinking about the tests caused me to think about the behaviour we're testing.

tests/test_27_optimise_random.py
```diff
@@ -0,0 +1,36 @@
...
+    def test_stopper_ok_1(self, stopper):
+        assert stopper.ok(1) is True
+        assert stopper.record == [1]
+
+    def test_stopper_ok_2(self, stopper):
+        assert stopper.ok(1) is True
+        assert stopper.ok(1.02) is True
+        assert stopper.record == [1, 1.02]
+
+    def test_stopper_ok_3_no_stop(self, stopper):
```
mfitz, line 27: This test is verifying that the stopper maintains a list of size `horizon`. But it's also testing the logic around the stop/don't stop decision, so there are two different things that can cause this test to fail.

If we care about how the record is maintained, I'd consider splitting that out into a separate test; something like:

```python
def test_pops_oldest_value_when_record_exceeds_horizon_length(self, stopper):
    for i in [1, 2, 3, 4]:
        stopper.ok(i)
    assert stopper.record == [2, 3, 4]
```

tests/test_27_optimise_random.py
brynpickering: Three reasons, although I'm not strongly attached to them:

1. more clear grouping of tests for parts of modules. Here, All my unit tests associated with the `Stopper` class are grouped within the test class.
2. fixtures defined inside a class are isolated from elsewhere in the test module, allowing name duplication if desired.
3. running a subset of tests within a module is easier, e.g. `pytest tests/test_27_optimise_random.py:TestStopper` would run all the tests in that class.

mfitz (changes requested):
Good to see more test coverage 👍 

I've made some minor comments about some of the tests, and also polluted those slightly with ponderings about the code under test, which is not part of your PR but is related to the new tests.

tests/test_27_optimise_random.py
brynpickering: Looking at how stopper gets used, the records should be ordered as they get added every time there is a positive increment in `score` (see lines 26:32 in the same file). To be honest, the implementation seems excessively complex for what it is doing, but I don't think it's worth the effort to refactor.

mfitz (approved):

PR #208, opened 2023-07-03
Update all docstrings to Google style

brynpickering:
The docstrings were following various styles, sometimes ReST, sometimes numpydoc, sometimes Google. Sometimes they were a francenstein's monster of these three, which no automatic parser can hope to deal with.

I have updated all existing docstrings to the Google style, which I believe to be the most readable of the options. 
It also allows for our documentation to parse `Note: ...` and `Tips: ...` as highlighted boxes as described [here](https://mkdocstrings.github.io/python/usage/docstrings/google/#docstring).

In the process of updating, I've tried to add as many type hints as I could. However in `pam/samples/tour.py`, there were instances where I failed to work out what an input argument's type was expected to be. @Theodore-Chatziioannou, can you help with annotating `add_return_origin`, `add_tour_activity`, `add_tour_leg`, `apply`?

I have also updated some `base` classes to actual python [Abstract Base Classes](https://docs.python.org/3/library/abc.html) since it made documenting them and their children easier. Based on the tests that were supposed to catch `NotImplementedError`s, I would say these classes *should* have always been ABCs.


Theodore-Chatziioannou:
> rk out what an input argument's type was expected to be. @Theodore-Chatziioannou, can you help with annotating `add_return_origin`, `add_tour_activity`, `add_tour_leg`, `apply`?
> 

@divyasharma-arup

fredshone (approved):
Wowzas. Suggest you merge and then we fill any gaps after. There are parts of pam that are used for "internal research" purposes. I do think that it is good to pressure such code to rise into shape, but that is beyond scope of single PR.

PR #210, opened 2023-07-04
Update dependencies to work from py3.9 to py3.11

brynpickering:
PAM should now work on any version from python 3.9 to 3.11, with the CI updated to test the extremes of this range.
It would also be possible to test with some dependencies set to lower bounds to ensure that breaking changes don't creep in (e.g., I've tested locally with pandas=1.5, plotly=4, gdal=3.5 and everything passes, just with some deprecation warnings).

I've put explicit warning ignores for persistent deprecation warnings that are the fault of dependencies, not us.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -3,40 +3,38 @@ channels:
   - conda-forge
   - default
 dependencies:
-    - click >= 8.1, < 9
+    - click < 9
     - gdal < 3.6
     - geopandas >= 0.13, < 0.14
-    - ipykernel >= 5.2, < 6
-    - lxml >= 4.6, < 5
-    - matplotlib >= 3.2, < 4
-    - pandas >= 1.3, < 2
-    - pip ~= 23.1
+    - ipykernel < 7
+    - lxml < 5
+    - matplotlib >= 3, < 4
+    - numpy >= 1, < 2
+    - pandas >= 1.5, < 3
+    - pip >= 23, < 24
     - pip:
       - s2sphere < 0.3
-    - plotly >= 4.10, < 5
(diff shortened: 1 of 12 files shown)


environment.yml
```diff
@@ -3,40 +3,38 @@ channels:
   - conda-forge
   - default
 dependencies:
-    - click >= 8.1, < 9
+    - click < 9
     - gdal < 3.6
```
brynpickering, line 7: this is kept as we found a windows issue when it was allowed to be higher

environment.yml
```diff
@@ -3,40 +3,38 @@ channels:
...
+    - xlrd >= 2, < 3
     # dev
-    - nbmake ~= 1.3
-    - pre-commit ~= 3.3
-    - pytest ~= 7.3
-    - pytest-cov ~= 4.1
-    - pytest-mock ~= 3.10
-    - pytest-timeout ~= 2.1
-    - pytest-xdist ~= 3.3
+    - nbmake < 2
```
brynpickering, line 27: Removed lower bounds for these development requirements as they shouldn't be allowed to influence the viability of the core requirements.

pam/planner/clustering.py
```diff
@@ -86,9 +86,7 @@ def fit(
         """
         if clustering_method == "agglomerative":
             model = AgglomerativeClustering(
-                n_clusters=n_clusters,
-                linkage=linkage,
-                affinity="precomputed",  # change argument to "metric" for sklearn version>=1.4
+                n_clusters=n_clusters, linkage=linkage, metric="precomputed"
```
brynpickering, line 89: fixes deprecation warning

PR #213, opened 2023-07-10
Time parse bugfix

fredshone:
Fixes #212  .

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/pam/utils.py
+++ b/pam/utils.py
@@ -14,12 +14,15 @@ from pam.variables import START_OF_DAY
 # according to gzip manpage
 DEFAULT_GZIP_COMPRESSION = 6
 
+
 def parse_time(time):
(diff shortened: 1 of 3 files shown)


pam/utils.py
```diff
@@ -90,8 +90,26 @@ def safe_strptime(mt):
     """Safely parse string into datetime, can cope with time strings in format hh:mm:ss
     if hh > 23 then adds a day.
     """
-    h, m, s = mt.split(":")
-    return START_OF_DAY + timedelta(hours=int(h), minutes=int(m), seconds=int(s))
+    units = mt.split(":")
```
brynpickering, line 93: This is almost entirely a duplication of `safe_strpdelta`, would be more maintainable to be simply `START_OF_DAY + safe_strpdelta(mt)`

PR #215, opened 2023-07-10
Vehicles api

fredshone:
WIP adds "MATSim vehicles" functionality.

Still has rough edges, needs documentation, changelog (ahem) and better coverage, but sharing for any high level complaints on the implementation which got out of control and needs feedback...

_I am aware I keep misspelling vehicles - it seems unavoidable_

# Aims:

At a super high level I am trying to facilitate two main use cases:

1. Create/add vehicles and write to matsim, eg (pseudo):
```
# create
population.vehicle_manager.add_type("default_car", VehicleType())
for hid, pid, person in population.people():
    person.vehicles = {"car" : Vehicle(pid, "default_car")}
    if some_condition(person):
        person.vehicles = {"car" : ElectricVehicle(pid, "default_car")}
# write
population.write_matsim(
    plans.xml,
    vehciles.xml,
    evs.xml
)
```
3. Read, edit and then write to matsim, eg (pseudo):
```
# read
population = read_matsim(
    plans.xml,
    vehciles.xml,
    evs.xml
)
# edit
population.vehicle_manager.add_type("small_car", VehicleType())
for hid, hh in population.households.items():
    if some_condition_a(hh):
        person = select_person(hh)
        person.vehicles = {"car" : Vehicle(pid, "small_car")}
# write
population.write_matsim(
    plans.xml,
    vehciles.xml,
    evs.xml
)
```

More generally I am adding the following
- I am making agent vehicles available as part of the core pam Population, Household and Person objects to allow easy access
- I am supporting vehicle types in order to have detailed control over veh specifications (such as size for freight)
- Multiple vehicles per agent (required for active mode extensions)
- Electric vehicles (I am mostly maintaining previous functionality)

# Challenge

The main cause of pain and complexity is that MATSim holds veh info in 3 places:

1. agent -> mode -> veh id. This is held as an agent attribute in plans.xml, eg:
```
<person id="census_0">
  <attributes>
    <attribute class="org.matsim.vehicles.PersonVehicles" name="vehicles">{"car": "census_0"}</attribute>
    <attribute class="java.lang.String" name="hid">census_0</attribute>
  </attributes>
...
```
2. veh id -> veh type. This is held in vehicles.xml, eg:
```
<vehicle id="census_0" type="small_car"/>
```
3. veh type -> veh attributes (eg size). Also held in vehciles.xml, eg:
```
<vehicleType id="small_car"><description>personal_vehicle</description><capacity seats="4" standingRoomInPersons="0"/><length meter="7.5"/><width meter="1.0"/><passengerCarEquivalents pce="1.0"/><networkMode networkMode="car"/><flowEfficiencyFactor factor="1.0"/></vehicleType
```
4. veh id -> electric veh attributes. Held in third file evs.xml, eg:
```
<vehicle id="census_9" battery_capacity="100" initial_soc="60" charger_types="default" vehicle_type="small_car"/
```

Because I wanted to allow veh editing at the agent level (eg `Person.vehicles["car"] == Vehicle(...)`), I felt there were two options:

A) A vehicle factory/manager that has control over both the agent population and veh population:
```
Population.add_veh_to(hid, pid, Vehicle)
```
This would maintain consistency betwen vehicles and agents owning vehicles.

B) Move all vehicles into Population during read and return prior to write:
```
Population[hid][pid].add_vehicle()
```

I prefered (B) because it's closer to "normy pam". I still create a VehiclesManager to represent vehicle types, look after read/write and help ensure consistency at read/write time. But this has ended up complex and maybe therefore fragile. Of particular note:

- @KasiaKoz 's OG vehicle api is very slightly broken. Mainly via the removal of the default `defaultVehicleType` for electric vehicles and the removal or the requirement that the veh id be the same as the person id. I had to do this to support multiple vehicles. Otherwise should work fine. 
- I am relying on users to maintain vehicle type key which are represented as a string. To add or edit a veh type spec, you have to access the manager, eg:
```
Population.vehicle_manager.add_type("TYPE_KEY", VehicleType(..))
```
Then to use it you specifiy that a Vehicle is of that type, eg:
```
Person(pid).vehicles["car"] = Vehicle("veh_id", "TYPE_KEY")
```
This is probably confusing and although generally I can clean up names and interfaces, there is still a string being used as a key. Perhaps I could point each vehicle at a VehicleType? But then we run into accidental edits.
- Not intending to add documentation. But I have added to the example notebook.
- I am aiming to make this a release

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


pam/core.py
```diff
@@ -176,47 +177,82 @@ def attributes(self, show: int = 10) -> dict:
...
+    def rebuild_vehicles_manager(self):
+        """
+        (Re)build veh population from agents.
+        """
+        self.vehicles_manager.clear_vehs()
+        self._build_vehicles_manager()
+
+    def _build_vehicles_manager(self):
+        """
+        Update veh population from agents.
```
brynpickering, line 199: not sure if you are mitigating the risk of misspelling "vehicle(s)" by using "veh(s)", but it does seem more readable to use the long format throughout.

pam/core.py
```diff
@@ -176,47 +177,82 @@ def attributes(self, show: int = 10) -> dict:
...
 
     @property
     def has_electric_vehicles(self):
-        return bool(list(self.electric_vehicles()))
+        return bool(self.evs())
+
+    def add_veh_type(self, vehicle_type: VehicleType):
+        self.vehicles_manager.add_type(vehicle_type)
+
+    def safe_add_veh_to_agent(self, hid: str, pid: str, mode: str, v: Vehicle) -> bool:
```
brynpickering, line 235: Any reason to prepend with "safe"? If you have an unsafe alternative method (I don't see one), it would be better to have `safe` as a boolean input arg and merege the two methods together.

pam/core.py
```diff
@@ -176,47 +177,82 @@ def attributes(self, show: int = 10) -> dict:
...
+    def vehicles_manager(self):
+        v_types = {p.vehicles.vehicle_type for _, _, p in self.people() if p.vehicles is not None}
         for vt in v_types:
             yield vt
 
     def electric_vehicle_charger_types(self):
         chargers = set()
-        for v in self.electric_vehicles():
+        for _, _, _, v in self.evs():
             chargers |= set(v.charger_types.split(","))
```
brynpickering, line 256:
```suggestion
            chargers.update(v.charger_types.split(","))
```

pam/core.py
```diff
@@ -924,14 +965,15 @@ def __init__(
         pid,
         freq=None,
         attributes={},
+        vehicles: dict[str:Vehicle] = {},
```
brynpickering, line 968:
```suggestion
        vehicles: dict[str, Vehicle] = {},
```

pam/core.py
```diff
@@ -924,14 +965,15 @@ def __init__(
...
         home_location: Optional[Location] = None,
         home_area=None,
         home_loc=None,
-        vehicle: Vehicle = None,
     ):
         self.pid = pid
         self.person_freq = freq
         self.attributes = attributes
+        self.vehicles = vehicles
         if home_location is not None:
```
brynpickering, line 977: I would probably do this with the args that default to empty dicts too. Generally a bad idea to have mutable objects as default values.

pam/read/matsim.py
```diff
@@ -126,9 +127,9 @@ def read_matsim(
 def stream_matsim_persons(
     plans_path: str,
     attributes: dict = {},
-    vehicles: dict = {},
+    vehicles_manager: Optional[VehicleManager] = VehicleManager(),
```
brynpickering, line 130: Only use `Optional` if `None` is a viable input.

tests/conftest.py
```diff
@@ -13,6 +14,35 @@
 from pam.variables import END_OF_DAY
 
 TEST_DATA_DIR = Path(__file__).parent / "test_data"
+FIXTURES_DIR = Path(__file__).parent.parent / "pam" / "fixtures"
```
brynpickering, line 17:
```suggestion
import importlib_resources
FIXTURES_DIR = importlib_resources.files("pam") / "fixtures"
```

pam/core.py
fredshone: I don't follow - `vehicles` is passed as an arg and defaults to {}?

brynpickering: Yes, nothing should default to `{}` in the function signature. Example of why:

``` python
def foo(bar, x = {}):
    x[bar] = 1
    return x

In [2]: foo(2)
Out[2]: {2: 1}

In [3]: foo(3)
Out[3]: {2: 1, 3: 1}  # expectation is {3: 1}
```

Better to do `x = None` and then handle the undefined case in the function.

fredshone:
ready for review again

pam/vehicles.py
```diff
@@ -0,0 +1,351 @@
...
+
+import pam.utils as utils
+from pam import PAMVehicleIdError
+
+# Vehicle classes to represent Vehicles based on MATSim DTD files:
+# https://www.matsim.org/files/dtd/vehicleDefinitions_v2.0.xsd
+# https://www.matsim.org/files/dtd/electric_vehicles_v1.dtd
+
+
+@dataclass(frozen=True)
```
brynpickering, line 15: Is this the only dataclass that needs `frozen` set? I see you added it for py3.11 compatibility - why do the others not need it?

pam/core.py
```diff
@@ -923,29 +977,33 @@ def __init__(
...
         home_area=None,
         home_loc=None,
-        vehicle: Vehicle = None,
     ):
         self.pid = pid
         self.person_freq = freq
         self.attributes = attributes
+
+        if vehicles is not None:
+            self.vehicles = vehicles
```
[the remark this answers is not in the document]
brynpickering: I would just make vehicles a property of the person which returns the vehicles dictionary. Ideally you would set `self._vehicles` in `__init__` and then return `self._vehicles` in the `vehicles` property. This way, you can't accidentally override `self.vehicles` elsewhere

pam/core.py
```diff
@@ -1041,6 +1100,15 @@ def num_legs(self):
             return len(list(self.legs))
         return 0
 
+    def vehicles(self):
```
brynpickering, line 1103: replace with property that returns `self._vehicles`

pam/core.py
```diff
@@ -1041,6 +1100,15 @@ def num_legs(self):
...
+        for mode, veh in self.vehicles.items():
+            yield mode, veh
+
+    def evs(self):
```
brynpickering, line 1107: replace with property that does the same thing, but returns a dictionary or an iterator of one

brynpickering: I would prefer a direct dicitonary so that it acts in the same way as `Person.vehicles`

pam/core.py
```diff
@@ -177,49 +177,98 @@ def attributes(self, show: int = 10) -> dict:
...
+        self.vehicles_manager.vehicles.update(vehs)
+        if not self.vehicles_manager.is_consistent():
+            raise UserWarning("Failed consistency check refer to logs.")
+
+    def vehicle_types(self) -> Iterator[str, VehicleType]:
+        """
+        Yields:
+            Iterator[str, VehicleType]: Iterator of (vehicle type id, vehicle type data).
+        """
+        for k, v in self.vehicles_manager.types():
```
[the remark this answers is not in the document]
brynpickering: There seems no getting around using `yield` for spitting out tuples from nested dicts (as is done elsewhere in PAM, inc. in core), but for flat dictionaries, `iter(dict_name.items())` would be better.

pam/core.py
```diff
@@ -177,49 +177,98 @@ def attributes(self, show: int = 10) -> dict:
...
-        for v in self.electric_vehicles():
-            chargers |= set(v.charger_types.split(","))
-        return chargers
+        Args:
+            type_id (str): Vehicle type uid.
+            vehicle_type (VehicleType): Vehicle type to be added.
+        """
+        self.vehicles_manager.add_type(type_id, vehicle_type)
+
+    def add_veh_to_agent(self, hid: str, pid: str, mode: str, vehicle: Vehicle):
```
brynpickering, line 255: Is there a danger to having the vehicles manager only available in the `Population` class? A user could add a vehicle to a `Person` directly that has an unknown type. In fact, that is what you're doing in the example notebook, rather than using this method.

I'd either have this as your only method to add a vehicle to an agent (hiding the `Person.vehicles` dictionary from public view), or propagate the vehicles manager to Households and Persons when they are added to the population, so that when you add a vehicle to a Person, it checks its validity within the vehicle manager.

brynpickering: Example drawing on your notebook:

This passes, but is inconsistent:
``` python
pop.vehicles_manager.add_type("small_car", vehicles.VehicleType())
pop.vehicles_manager.add_type("large_car", vehicles.VehicleType(length=15, width=2))
for hid, pid, p in pop.people():
    p.vehicles = {
            "car": vehicles.ElectricVehicle(pid, type_id="medium_car", battery_capacity=100)
        }
```

This fails as expected:
``` python
pop.vehicles_manager.add_type("small_car", vehicles.VehicleType())
pop.vehicles_manager.add_type("large_car", vehicles.VehicleType(length=15, width=2))
for hid, pid, p in pop.people():
    pop.add_veh_to_agent(hid, pid, "car", vehicles.ElectricVehicle(pid, type_id="medium_car", battery_capacity=100))
```

examples/data/example_data/evs.xml
brynpickering, line 1: Add these to a subdirectory specific to the vehicles example, otherwise it's unclear when looking in the data directory that they are connected to each other and not to anything else in the `example_data` subdir

pam/write/matsim.py
```diff
@@ -16,15 +17,15 @@
...
 
 def write_matsim(
-    population: Population,
-    plans_path: str,
-    attributes_path: Optional[str] = None,
-    vehicles_dir: Optional[str] = None,
-    version: Optional[int] = None,
+    population,
+    plans_path: Union[Path, str],
+    attributes_path: Union[Path, str, None] = None,
```
brynpickering, line 25:
```suggestion
    attributes_path: Optional[Union[Path, str]] = None,
```
Although the result is the same, this is clearer as a hint, saying that useful inputs are `Path`/`str` and `None` is a placeholder to indicate you want to ignore that arg.

brynpickering: Same with the below args

pam/write/matsim.py
```diff
@@ -16,15 +17,15 @@
...
+    vehs_path: Union[Path, str, None] = None,
+    evs_path: Union[Path, str, None] = None,
+    version: int = None,
```
brynpickering, line 28: It should be wrapped in `Optional` since it defaults to None...
Since it is legacy, can you add a deprecation warning that is thrown when `version` is not None, telling the user that `version` will have no effect and the arg will be removed entirely in e.g. version `0.3.0`

pam/write/matsim.py
```diff
@@ -179,17 +181,16 @@ def create_person_element(pid, person, keep_non_selected: bool = False):
     person_xml = et.Element("person", {"id": str(pid)})
 
     attributes = et.SubElement(person_xml, "attributes", {})
+    if person.vehicles:
+        attribute = et.SubElement(
+            attributes,
+            "attribute",
+            {"class": "org.matsim.vehicles.PersonVehicles", "name": str("vehicles")},
```
brynpickering, line 188: why `str("vehicles")` and not `"vehicles"`

brynpickering (changes requested):

pam/vehicles.py
fredshone: I think this occurs because the dataclass is used as a default in the `VehicleType` class:

```
@dataclass
class VehicleType:
    capacity: CapacityType = CapacityType()
```

this risks the classic (you reminded me last week) mutable default bug where the default value gets changed by each instatiation. I think the dataclasses mod is picking this up.

Rather than freeze I can also do this to keep it mutable:

```
@dataclass
class VehicleType:
    capacity: CapacityType = field(default_factory=lambda: CapacityType())
```

From [docs](https://docs.python.org/3/library/dataclasses.html#dataclasses.field):

> default_factory: If provided, it must be a zero-argument callable that will be called when a default value is needed for this field. Among other purposes, this can be used to specify fields with mutable default values, as discussed below. It is an error to specify both default and default_factory.

For some reason I cannot call `CapacityType()` directly  (I though this would be a "callable") - so i add the lambda.

pam/core.py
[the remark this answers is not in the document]
fredshone: removed the iterator (i agree with your suggestion) but pam elsewhere just exposes mutable data and i don't want to get into get/set methods everywhere

pam/vehicles.py
brynpickering: Would it work as just `CapacityType` (no brackets)?

pam/core.py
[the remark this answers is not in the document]
fredshone: Thanks for this - I cleaned out the unnecessary ones and replaced with @property where possible. Otherwise I am going to expect people to interact with the attributes directly. This is generally the case throughout.

pam/write/matsim.py
fredshone: probably a mental breakdown

pam/core.py
fredshone: yeah this is the crux of the PR and i agree with comment. But... I am still keen to maintain a simple python object based representation that allows for basic use cases (no matsim) that don't care about vehicle types in an especially systematic way.

So for now I've added a new `Population.check_vehicles()` method and included in the example. I will refine docs and await a third part to decide :)

brynpickering:
You seem to have deleted the data that the example notebook needs to run, hence the failing tests.

pam/core.py
brynpickering: OK, that makes sense. Minor point if keeping `add_veh_to_agent`: `agent` isn't really used in PAM, so should it be `add_vehicle_to_person`?

fredshone:
huzzah

pam/vehicles.py
```diff
@@ -0,0 +1,433 @@
...
+
+    def __init__(self) -> None:
+        self.veh_types = {}
+        self.vehicles = {}
+
+    def add_type(self, k: str, vehicle_type: VehicleType) -> None:
+        """Add vehicle type to manager.
+
+        Args:
+            k (str): Vehicle type key.
```
KasiaKoz, line 192: I feel like we should add a `type_id` to `VehicleType` class to use instead of `k` here, otherwise the link to the type name is only in the manager.. unless you want to be able to use the same `VehicleType` with different type keys/names?...

fredshone: I have some struggles and doubts with this throughout:

(i) It is convenient to keep the id as a hash key, so that I can access cheaply.
But also (ii) the Vehicles class needs an "internal" veh_id so that it can be held as an agent attribute (eg `{"car": Vehicle(...)}`.
But then (iii) I don't like the duplication of the id, eg `{"type_0": VehicleType("type_0")}`.

So to summarise, I actually prefer this key value representation: `{"type0": VehicleType()}`. But I have to break it for vehicles because of the broader requirements of what i implemented.

Option 1: keep as is, eg `{"type0": VehicleType(), type1": VehicleType(), ...}`.
Option 2: move to list, eg `[VehicleType("type0"), VehicleType("type1"), ...]`, this removes key but slows lookups.
Option 3: duplicate, eg `{"type0": VehicleType("type0"), type1": VehicleType("type1"), ...}`.

3 is not much work but adds duplication and the possibility of inconsistency. ie what if someone does this: `VehicleManager.veh_types["typO"] = VehicleType("typE")`.

Same snaffu exists for Vehicles, but is less likely because the link is generally established in the population where the key is the mode. Whereas types only exist in the manager.

The other reason I had for using keys is that the Vehicles can then reference the key of the VehcileType, rather than the VehicleType itself. This seemed safer to me, because it would make it harder for someone changing the values of a type (such as the capacity) for one person's vehicle, accidentally changing it for everyone else.

Gnerally I realise this happens because we are storing VehcileType data for each Vehcile as a reference rather than data belonging to the Vehcile itself. But I am happy to do this as it reduces wasted memory and is consistent with MATSim.

Not sure if any of these helps. Let me know what you think now?

brynpickering: There's a lot of confusion introduced by the various ways you can assign things in this PR. I know it is there to provide flexibility, but might still benefit from tightening the ways in which assignment is possible. I wouldn't allow access to `VehicleManager.veh_types`. This would be a private property of the manager that can only be added to by calling `add_type`. This way, the internal data structure (dicts with `ID: VehicleType`) can exist with duplication of type ID, but the user doesn't see it. They can then add a type by calling `add_type(VehicleType(type_name))`. Would that work or am I being naive about where the type is assigned and accessed in the code?

fredshone: ok - I have gone for option 3: `{"type0": VehicleType("type0"), type1": VehicleType("type1"), ...}`

and as Bryn suggests, I've made the data more "private", such that there is no direct interaction with the vehicles manager encouraged. See example notebook for "proof".

KasiaKoz: I like that option and hiding the vehicle manager. 
Sorry for not replying earlier, just wanted to add, in my implementation I chose to make the objects immutable with `@dataclass(frozen=True)` which makes it a bit safer (also from real life perspective vehicle specs don't really change, you might get a new vehicle with different spec) but maybe overkill and I think you had reasons for changing that which I can't recall right now

fredshone: I resolve this

KasiaKoz (approved):

fredshone:
@brynpickering I think i need you nod - pretty sure I've incorporated all your feedback...

brynpickering (approved):

PR #216, opened 2023-07-11
Adds utils docstrings and typing

fredshone:
Redo of fix #212 which i [forced](https://github.com/arup-group/pam/pull/213). As penance for previous bad behaviour I've done the pam.utils typing and docstrings, plus change as per [this](https://github.com/arup-group/pam/pull/213#discussion_r1258416744) comment.

I - afterwards - see there is already a [PR](https://github.com/arup-group/pam/pull/208) for docstrings that includes utils. Happy to sit on this until after that is merged. 🤦🏻 

re parameterising - I am happy not to parameterise the tests as they are one liners in any case.

## Checklist

- [x] CHANGELOG updated - I do not think required
- [x] Documentation updated


brynpickering:
Turns out that `hh:mm:ss` in the docs is rendered as: 
image

brynpickering (approved):
Nice work, you are duly forgiven.

I made some minor changes when merging changes from the docstring update PR.

pam/utils.py
```diff
@@ -171,24 +184,50 @@ def safe_strpdelta(mt):
     raise UserWarning(f"Unrecognised timedelta format: {mt}")
```
brynpickering, line 184: Can you add a test to check that this is caught. Don't think there is one right now.

brynpickering (changes requested):

brynpickering (approved):

PR #217, opened 2023-07-13
Update utils and associated tests

brynpickering:
I noticed that your recently merged PR (#216) had tests that had too many assertions in one test. 

On checking that an error is caught, if you have several assertions inside the `with` statement, it will pass provided the first line raises the expected error.

I parametrized a bunch of tests in `test_00_utils.py` which I hope will show how easy it is to do so by default. 

I also added some tests to increase coverage and simplified some util functions that could rely on other functions in the same module, rather than have their own code.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/pam/utils.py
+++ b/pam/utils.py
@@ -19,7 +19,7 @@ DEFAULT_GZIP_COMPRESSION = 6
 def parse_time(time: Union[int, str]) -> datetime:
     """Generic parse time into datetime.
 
-    Itegers are assumed to be minutes.
+    Integers are assumed to be minutes.
     Strings are assumed to be datetime formatted as (%Y-%m-%d %H:%M:%S).
 
     Args:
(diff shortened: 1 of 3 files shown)


pam/utils.py
```diff
@@ -32,28 +32,26 @@ def parse_time(time: Union[int, str]) -> datetime:
         datetime: datetime
     """
 
-    if isinstance(time, int) or isinstance(time, np.int64):
+    if isinstance(time, (int, np.integer)) and not isinstance(time, bool):
```
brynpickering, line 35: note that `minutes_to_datetime` can handle float values (the remainder will be turned into seconds). If you *never* want this to be possible, float values should really be caught in `minutes_to_datetime` and lead to a raised error.

pam/utils.py
```diff
@@ -68,16 +66,16 @@ def datetime_string_to_datetime(string: str) -> datetime:
...
 
     Args:
-        minutes (int): minutes
+        minutes (float): minutes
 
     Returns:
         timedelta: timedelta
     """
-    return timedelta(minutes=minutes)
+    return timedelta(minutes=float(minutes))
```
brynpickering, line 78: timedelta doesn't like random numpy types (e.g., `np.int64`)

pam/utils.py
brynpickering: It's not about type safety so much as expected result safety. If you allow for only integers then your datetime will only have non-zero values in `year`, `hour`, `day`, `minute`. If you allow for floats (as is currently the case), then you will also have potential non-zero values in `second` which you may not expect when querying the object at other points in the code.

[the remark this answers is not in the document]
brynpickering: bools are also ints, but ints are not also bools. Depends on if you're ok with `True` == 1 minute or not.

pam/utils.py
```diff
@@ -32,28 +32,26 @@ def parse_time(time: Union[int, str]) -> datetime:
         datetime: datetime
     """
 
-    if isinstance(time, int) or isinstance(time, np.int64):
+    if isinstance(time, (int, np.integer)) and not isinstance(time, bool):
         return minutes_to_datetime(time)
-    if isinstance(time, str):
+    elif isinstance(time, str):
```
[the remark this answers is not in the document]
brynpickering: That's a new one on me. `elif` and `else` help avoid fall-throughs - cases where the same input is evaluated as `True` in multiple competing conditionals. Here, it's trivial since every conditional has a return statement in it, but if there is a variable assignment, you run the risk of accidental re-assignment if the if statement evaluates as `True` more than once. Here, If anything, we should avoid returns in the conditionals:

``` python
if isinstance(time, (int, np.integer)) and not isinstance(time, bool):
    dt = minutes_to_datetime(time)
elif isinstance(time, str):
    dt = datetime_string_to_datetime(time)
else:
    raise TypeError(
        f"Cannot parse {time} of type {type(time)} that is not int (assuming minutes) or str (%Y-%m-%d %H:%M:%S)"
    )

return dt
```

To me, it's more readable since you can follow the purpose of the function more clearly. Anyway, not something to get hung up on. So long as there is good test coverage, you can implement it how you like ;)

brynpickering:
RE `parse_time`, I think it's fine. We have some similar implementation dotted about in Calliope. Users can input some weird shit. For timeseries, though, we tend to be reasonably strict and say all CSVs must be the same datetime format and they have to specify that format in the config, if it deviates from the default `%Y-%m-%d %H:%M:%S`.

PR #219, opened 2023-07-14
adds optional research deps

### LINKED ISSUE #218: Pip install optional dependencies not working
### What happened?

`pip install ./pam[dev]` not working.

This is a zsh issue - fix incoming.

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Version

None

### Relevant log output

_No response_

fredshone:
Doc update fixes  #218.

zsh parses square brakets weirdly - so I've added `'` to relevant part of docs, eg `'./pam[dev]'`. I think zsh is common enough that we should do this rather than add additional docs for different shells. Also knowledgeable users can omit.

Also re-added the "research" optional dependencies (formerly called "planner"). I'm unclear on the real value of this if mamba is providing. But definately more pleasant for light venv or similar use. Documentation also bumped.

Longer term we might wanna spin out "research" modules into a new project. Not such a big deal now, but if we add more research leaning extensions - we probably don't need these to jump through the same QA as core pam. This seems especially plausible with base pam on conda. 

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/.github/workflows/build_pipeline.yml
+++ b/.github/workflows/build_pipeline.yml
@@ -133,7 +133,7 @@ jobs:
 
     - name: Send build success notification
       if: success()
-      uses: rtCamp/action-slack-notify@v2.0.0
+      uses: rtCamp/action-slack-notify@v2.2.0
       env:
         SLACK_MESSAGE: ${{ github.repository }} build ${{ github.run_number }} launched by ${{ github.actor }} has succeeded
         SLACK_TITLE: Build Success
@@ -144,7 +144,7 @@ jobs:
 
     - name: Send build failure notification
       if: failure()
-      uses: rtCamp/action-slack-notify@v2.0.0
+      uses: rtCamp/action-slack-notify@v2.2.0
       env:
         SLACK_COLOR: '#FF0000'
         SLACK_MESSAGE: ${{ github.repository }} build ${{ github.run_number }} launched by ${{ github.actor }} has failed
--- a/.github/workflows/daily-scheduled-ci.yml
+++ b/.github/workflows/daily-scheduled-ci.yml
@@ -38,7 +38,7 @@ jobs:
 
     - name: Send build success notification
       if: success()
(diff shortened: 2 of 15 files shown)


brynpickering (changes requested):
I'm generally against approaching package dependencies like this. Some projects do separate out optional and required dependencies, particularly in their pip requirements. However, the annoyance of working out what does and doesn't need installing when you hit a random import error isn't worth the few (ok, sometimes hundreds of) MB you save in your environment.

In PAM, the core code includes scripts that depend on `scikit-learn` and `python-levenshtein`. These need to be in the base package requirements. Since the conda package now depends on the same requirements file, it isn't possible to move it out for the sake of pip installs alone.

It would be possible to add notes across the documentation and the online package indexes as to what are optional depencies, but it just adds to confusion.

And ultimately, what is "optional"? If I am never going to load a package from excel, then `xlrd` is optional, if I'm never going to print summary reports, `prettytable` is optional. If I don't want to plot anything, `plotly` and `matplotlib` are optional (indeed, the memory footprint of these two and their dependencies is higher than `scikit-learn`'s). This is why I'd say that if we ever import a package in the source code, it *has* to be in the list of requirements.

Final point: `pip` and `mamba`(/`conda`) cache packages, so the footprint on disk of an environment is rarely as high as its contained packages, since you will invariably share some with other environments you have installed. This doesn't hold for docker containers, of course.

For a slim PAM package in future, "research leaning" code probably does want to move elsewhere. I would say the same for plotting code.

fredshone:
@brynpickering "research" optional deps removed. I appreciate your thoughts. Theo also explained that one rationale for making optional originally was to make it easier for windows peeps, who had trouble with sci-kit and levenstein. But mamba has made this concern redundant.

Minor doc changes still remain.

brynpickering (approved):
Confirmed as working with [fish shell](https://fishshell.com/) too

PR #222, opened 2023-07-18
fix for #221

### LINKED ISSUE #221: pam.activity.simplify_pt_trips is broken
### What happened?

pam.activity.simplify_pt_trips is broken assumes transit has mode "pt". This is no longer the case so method is broken.

- [X] macOS
- [ ] Windows
- [ ] Linux

### Version

0.2.5-dev

### Relevant log output

_No response_

fredshone:
Fixes #221.

Provides "pt" simplification based on removing "pt interaction" activities. Added coverage but am also going to test simplified plans in matsim. Will update.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/pam/activity.py
+++ b/pam/activity.py
@@ -61,6 +61,64 @@ class Plan:
             if isinstance(p, Leg):
                 yield p
 
+    def simplify_pt_trips(self, skip=["pt interaction", "pt_interaction"]) -> None:
+        """Remove pt interaction events and simplify legs to single leg with mode.
+
+        pt interaction events result from complex matsim plans.
+
+        """
+        if set(skip).intersection(self.activity_classes):
+            self.day = list(self.tripify(skip=skip))
+
+    def tripify(
+        self, skip: list[str] = ["pt interaction", "pt_interaction"]
+    ) -> Iterator[PlanComponent]:
+        """Yield plan trips and activities based on ignoring certain activities.
+
+        Args:
+          ignore (list[str], optional): activities to ignore. Defaults to ["pt interaction", "pt_interaction"].
+
+        Returns:
+            Iterator: PlanComponents
+        """
+        if self.day:
+            seq = 0
+            modes = {}
+            start_location = self.day[0].location
+            start_time = self.day[0].end_time
+            distance = 0
+            for component in self:
+                if isinstance(component, Leg):
+                    modes[component.mode] = modes.get(component.mode, 0) + component.distance
+                    distance += component.distance
+                    attributes = component.attributes  # trips collect attributes from last leg
+                if isinstance(component, Activity):
+                    if component.act not in skip:
+                        while True:
+                            if modes:
+                                yield Leg(
+                                    seq=seq,
+                                    mode=max(modes, key=modes.get),
+                                    start_area=start_location.area,
+                                    end_area=component.location.area,
+                                    start_link=start_location.link,
+                                    end_link=component.location.link,
+                                    start_loc=start_location.loc,
+                                    end_loc=component.location.loc,
+                                    start_time=start_time,
+                                    end_time=component.start_time,
+                                    distance=distance,
+                                    purp=component.act,
+                                    attributes=attributes,
+                                )
+                            yield component
+                            break
+                        modes = {}
+                        start_location = component.location
+                        start_time = component.end_time
+                        distance = 0
+                        seq += 1
+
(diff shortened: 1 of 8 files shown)


pam/activity.py
```diff
@@ -61,6 +61,64 @@ def legs(self):
             if isinstance(p, Leg):
                 yield p
 
+    def simplify_pt_trips(self, skip=["pt interaction", "pt_interaction"]) -> None:
+        """Remove pt interaction events and simplify legs to single leg with mode.
+
+        pt interaction events result from complex matsim plans.
```
brynpickering, line 67: Add Args section as with `tripify`

pam/activity.py
```diff
@@ -61,6 +61,64 @@ def legs(self):
...
+                                )
+                            yield component
+                            break
+                        modes = {}
+                        start_location = component.location
+                        start_time = component.end_time
+                        distance = 0
+                        seq += 1
+
     def trips(self, ignore: list[str] = ["pt interaction", "pt_interaction"]) -> str:
```
brynpickering, line 122: `ignore` -> `skip` (or vice versa) to align argument names between methods?

pam/read/matsim.py
```diff
@@ -449,7 +449,20 @@ def get_attributes_from_person(elem):
     ident = elem.xpath("@id")[0]
     attributes = {}
     for attr in elem.xpath("./attributes/attribute"):
-        attributes[attr.get("name")] = attr.text
+        data = attr.get("class")
```
brynpickering, line 452: In the if statements, assign the value to a variable, e.g. `typed_attr`, then assign that to `attributes[attr.get("name")]` at the end. This reduces lines that need to change in future if `attributes[attr.get("name")]` changes for any reason.

pam/read/matsim.py
```diff
@@ -449,7 +449,20 @@ def get_attributes_from_person(elem):
...
+        if data == "java.lang.String":
+            attributes[attr.get("name")] = attr.text
+        elif data == "java.lang.Boolean":
+            attributes[attr.get("name")] = attr.text == "true"
```
brynpickering, line 456: can it strictly only ever be "true"? not "TRUE" or "True"? If you're unsure, then do `attr.text.lower()`

tests/test_02_activity_trips.py
```diff
@@ -0,0 +1,78 @@
...
+        ),
+    ],
+)
+def test_iter_trips(day, expected):
+    plan = Plan()
+    plan.day = day
+    assert list(plan.trips()) == expected
+
+
+@pytest.mark.parametrize(
```
brynpickering, line 40: Split in two: ones where you don't expect a change in the data and ones where you do. Then you can reduce reproduction of code in the test where you don't expect a change (because `day == expected`)

brynpickering (changes requested):
A few minor comments.

On the implementation, `tripify` seems very specific to public transport interaction legs/activities since you carefully remove the leg leading up to a `pt interaction` that comes before a public transport leg, then remove the leg following a `pt interaction` that comes after a public transport leg (I could only really infer this from the tests), e.g.:

``` python 
[                
    Activity(act="home", loc=Point(0, 0)),
    Leg(mode="walk", distance=100, start_loc=Point(0, 0), end_loc=Point(100, 0)),
    Activity(act="pt interaction", loc=Point(100, 0)),
    Leg(mode="bus", distance=1000, start_loc=Point(100, 0), end_loc=Point(100, 1000)),
    Activity(act="pt interaction", loc=Point(100, 1000)),
    Leg(mode="walk", distance=20, start_loc=Point(100, 1000), end_loc=Point(0, 1000)),
    Activity(act="work", loc=Point(0, 1000)),
]
```

becomes

``` python
[
    Activity(act="home", loc=Point(0, 0)),
    Leg(mode="bus", distance=1120, start_loc=Point(0, 0), end_loc=Point(0, 1000)),
    Activity(act="work", loc=Point(0, 1000)),
]
```

However, the docstring doesn't explain that this is the expected result, instead of, say:

``` python
[                
    Activity(act="home", loc=Point(0, 0)),
    Leg(mode="walk", distance=1120, start_loc=Point(0, 0), end_loc=Point(0, 1000)),
    Activity(act="work", loc=Point(0, 1000)),
]
```
If it's specific to removing public transport interactions, then I would make that much more obvious.

fredshone:
Good feedback thanks @brynpickering - I have addressed all as you suggest.

PR #230, opened 2023-07-27
Update install instructions and docker entry point

### LINKED ISSUE #228: Docker container entry point should be `python` not `ipython`
### What happened?

I tried using the generated PAM Docker image to run a job on AWS and realised that the entry point for the container should not be `ipython`, but rather `python`. I initially set the entry point to `ipython` as I assumed most people would interact with the container interactively, but that was probably only true when installing PAM on Windows devices was a massive pain. Now, most people are going to run python scripts as commands on the Docker container from an e.g. AWS job, so `python` makes more sense.

In the meantime, it is possible to get the same result as `python -m my_script ....` by calling `ipython -- my_script ....`

- [X] macOS
- [ ] Windows
- [X] Linux

### Version

0.2.5-dev

### Relevant log output

_No response_

### LINKED ISSUE #229: Make dev install instructions more prominent
### Description

Most PAM users are also developers, so the developer / contributer installation instructions should be more clear in the README and in the documentation. The README currently only gives the _user_ installation instructions, leading to an environment being installed without the expected testing/docs dependencies.

### Related links

- https://arup-group.github.io/pam/develop/installation/
- https://arup-group.github.io/pam/develop/contributing/coding/
- https://github.com/arup-group/pam/blob/main/README.md

### Version

0.2

### Proposed change

Add this code snippet (which includes the dev requirements) to the installation instructions and the README under an e.g., `as a developer` subheading:

``` shell
git clone git@github.com:arup-group/pam.git
cd pam
mamba create -n pam -c city-modelling-lab --file requirements/base.txt --file requirements/dev.txt
mamba activate pam
pip install --no-deps -e .
```

brynpickering:
Fixes #228
Fixes #229

This cleans up some parts of the documentation surrounding installation. E.g., now the short-form install instructions are:

image

I've also slipped in updating the dockerfile, so that the entry point to the image is plain bash, instead of `python`.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,4 +6,4 @@ ARG MAMBA_DOCKERFILE_ACTIVATE=1
 
 RUN pip install --no-deps -e .
 
-ENTRYPOINT ["/usr/local/bin/_entrypoint.sh", "ipython"]
+ENTRYPOINT ["/usr/local/bin/_entrypoint.sh"]
--- a/README.md
+++ b/README.md
@@ -49,25 +49,36 @@ supply or utility demand.
 The PAM package is not yet indexed online.
 To install it, we recommend the following:
 
+### As a user
+<!--- --8<-- [start:docs-install-user] -->
 ``` shell
 git clone git@github.com:arup-group/pam.git
 cd pam
-mamba create -n pam -c city-modelling-lab --file requirements/base.txt
+mamba create -n pam -c conda-forge -c city-modelling-lab --file requirements/base.txt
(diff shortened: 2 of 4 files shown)


fredshone (commented):
I think that talking about indexing in the first line will be unhelpful to 90% of first time users. I think you also need to add a link to mamba. At lest until it is on equal footing with conda.

fredshone (approved):

PR #231, opened 2023-07-28
Freight route optimisation

val-ismaili:
Based on the route optimisation challenge [here](https://github.com/arup-group/activity-synthesis/blob/main/location_choice/example_freight_location.ipynb) showing the approximated `greedy_tsp` as the most effective balance between optimised routing and computing time, this has been implemented within `tour.py` for stop sequencing of freight delivery locations.

During implementation, it was found that the existing code allowed for duplicate sampling of delivery locations within the same sequence as well as the possibility for the depot being sampled as a delivery location (if the facility has both tags). Both these issues have been fixed within `sample_destinations()` in `tour.py`. Associated test has been added with `test_unique_stops()` in `test_09_samplers_tour.py`. This was discovered when testing on toy data which resulted in occasional duplicate sampling of delivery locations resulting in a distance_matrix with zero distances resulting in an incomplete graph - test is added for this too under `test_distance_matrix_is_complete`.

## Checklist
- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/pam/samplers/tour.py
+++ b/pam/samplers/tour.py
@@ -319,6 +319,7 @@ class TourPlanner:
         o_loc = self.facility_sampler.sample(self.o_zone, self.o_activity)
 
         d_seq = []
+        sampled_d_facilities = []
 
         for j in range(self.stops):
             # If threshold matrix is none, sample a random d_zone, else select a d_zone within threshold value
@@ -335,7 +336,16 @@ class TourPlanner:
             d_facility = self.facility_sampler.sample(d_zone, self.d_activity)
 
             # prevent the depot from being sampled as a delivery
-            while d_facility == o_loc:
+            while d_facility == o_loc or d_facility in sampled_d_facilities:
+                if self.threshold_matrix is None:
+                    d_zone = FrequencySampler(self.d_dist.index, self.d_dist[self.d_freq]).sample()
+                else:
+                    d_zone = FrequencySampler(
+                        dist=self.d_dist,
+                        freq=self.d_freq,
+                        threshold_matrix=self.threshold_matrix.loc[self.o_zone],
+                        threshold_value=self.threshold_value,
+                    ).threshold_sample()
                 d_facility = self.facility_sampler.sample(d_zone, self.d_activity)
 
             # append to a dictionary to sequence destinations
--- a/pam/samplers/tour.py
+++ b/pam/samplers/tour.py
@@ -279,7 +279,7 @@ class TourPlanner:
         hour: int,
         minute: int,
         o_zone: str,
-        d_dist: Union[Iterable, pd.DataFrame],
+        d_dist: pd.DataFrame,
         d_freq: Union[str, Iterable],
         facility_sampler: FacilitySampler,
         activity_params: dict[str, str],
--- a/pam/samplers/tour.py
+++ b/pam/samplers/tour.py
@@ -269,7 +269,7 @@ class DurationEstimator:
 class TourPlanner:
     """Object for agents to efficiently plan their tours by sequencing stops and adding activities and legs.
 
-    The TourPlanner optimizes the sequence of stops using a Greedy Travelling Salesman Problem (TSP) algorithem based on Eucledian distances between sampled stops.
+    The TourPlanner optimises the sequence of stops using a Greedy Travelling Salesman Problem (TSP) algorithm based on Eucledian distances between sampled stops.
     It takes into account origin and destination zones, facility distributions, and other relevant parameters to build a tour plan for agents.
     """
 
--- a/pam/samplers/tour.py
+++ b/pam/samplers/tour.py
@@ -326,17 +326,7 @@ class TourPlanner:
 
         return d_zone
 
-    def sequence_stops(self) -> tuple[list, list, list]:
-        """Creates a sequence for a number of stops. Sequence is determined by distance from origin.
-
-        TODO - Method to sequence stops with different logic (i.e, minimise distance between stops).
-
-        Returns:
-          tuple[list, list, list]: (o_loc, d_zones, d_locs).
-
-        """
-        o_loc = self.facility_sampler.sample(self.o_zone, self.o_activity)
-
+    def sample_destinations(self, o_loc):
         d_seq = []
         sampled_d_facilities = []
 
@@ -363,7 +353,9 @@ class TourPlanner:
                     "distance": ActivityDuration().model_distance(o_loc, d_facility),
                 }
             )
+        return d_seq
 
+    def create_distance_matrix(self, o_loc, d_seq):
         # extact o_loc coordinates into array
         o_location = np.array([[o_loc.x, o_loc.y]])
 
@@ -377,16 +369,39 @@ class TourPlanner:
         # approximated from reduce compute time
         locs = np.concatenate([o_location, d_locations], 0)
         dist_matrix = distance_matrix(locs, locs)
-        # print(d_seq)
-        # print(dist_matrix)
+
+        return dist_matrix
+
+    def approx_greedy_tsp(self, dist_matrix):
         distance_graph = nx.from_numpy_array(dist_matrix)
         seq = nx.algorithms.approximation.greedy_tsp(distance_graph, source=0)
 
+        return seq
+
+    def reorder_deliveries(self, d_seq, seq):
         # use `seq` to re-order `d_locs` into an ordered list of dictionaries
         # remove o_loc (first and last stop) from sequence & adjust sequence range
         seq = seq[1:-1]
         seq = [x - 1 for x in seq]
         d_seq = [d_seq[order] for order in seq]
+
+        return d_seq
+
+    def sequence_stops(self) -> tuple[list, list, list]:
+        """Creates a sequence for a number of stops. Sequence is determined by approximated greedy TSP
+
+        Returns:
+          tuple[list, list, list]: (o_loc, d_zones, d_locs).
+
+        """
+        o_loc = self.facility_sampler.sample(self.o_zone, self.o_activity)
+
+        d_seq = TourPlanner.sample_destinations(self, o_loc)
+        dist_matrix = TourPlanner.create_distance_matrix(self, o_loc, d_seq)
+
+        seq = TourPlanner.approx_greedy_tsp(self, dist_matrix)
+
+        d_seq = TourPlanner.reorder_deliveries(self, d_seq, seq)
 
         # sort distance: furthest facility to closest facility to origin facility. The final stop should be closest to origin.
         d_zones = [item.get("destination_zone") for item in d_seq]
--- a/pam/samplers/tour.py
+++ b/pam/samplers/tour.py
@@ -3,12 +3,12 @@ import warnings
 from typing import Any, Iterable, Optional, Union
 
 import geopandas as gp
+import networkx as nx
 import numpy as np
 import pandas as pd
 from matplotlib import pyplot as plt
-from shapely.geometry import Point
 from scipy.spatial import distance_matrix
-import networkx as nx
+from shapely.geometry import Point
 
 from pam.activity import Activity, Leg
 from pam.samplers.facility import FacilitySampler
@@ -334,6 +334,10 @@ class TourPlanner:
             # once d_zone is selected, select a specific point location for d_activity
             d_facility = self.facility_sampler.sample(d_zone, self.d_activity)
 
+            # prevent the depot from being sampled as a delivery
+            while d_facility == o_loc:
+                d_facility = self.facility_sampler.sample(d_zone, self.d_activity)
+
             # append to a dictionary to sequence destinations
             d_seq.append(
                 {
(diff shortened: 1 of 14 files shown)


pam/samplers/tour.py
```diff
@@ -305,33 +307,43 @@ def __init__(
...
+        """Samples a delivery zone (d_zone) as a string, dependent on the presence of a threshold matrix
 
         Returns:
-          tuple[list, list, list]: (o_loc, d_zones, d_locs).
+            str: d_zone
 
         """
-        o_loc = self.facility_sampler.sample(self.o_zone, self.o_activity)
+        if self.threshold_matrix is None:
+            d_zone = FrequencySampler(self.d_dist.index, self.d_dist[self.d_freq]).sample()
```
brynpickering, line 318: self.d_dist can supposedly be an `Iterable` (see class init type hints), in which case it wouldn't have an `index`, on a pandas dataframe would. Do you only ever expect `d_dist` to be a dataframe now?

pam/samplers/tour.py
```diff
@@ -305,33 +307,43 @@ def __init__(
         self.o_activity = activity_params["o_activity"]
         self.d_activity = activity_params["d_activity"]
 
-    def sequence_stops(self) -> tuple[list, list, list]:
-        """Creates a sequence for a number of stops. Sequence is determined by distance from origin.
-
-        TODO - Method to sequence stops with different logic (i.e, minimise distance between stops).
+    def d_zone_sample_choice(self) -> str:
+        """Samples a delivery zone (d_zone) as a string, dependent on the presence of a threshold matrix
```
brynpickering, line 311: delivery -> destination? In the class docstring it is `destination` while here it is `delivery`, which is a bit confusing as to whether `d` now means something different

pam/samplers/tour.py
```diff
@@ -305,33 +307,43 @@ def __init__(
...
+                threshold_value=self.threshold_value,
+            ).threshold_sample()
+
+        return d_zone
 
+    def sample_destinations(self, o_loc):
         d_seq = []
+        sampled_d_facilities = []
 
         for j in range(self.stops):
```
brynpickering, line 333: while we're here (I know you haven't edited this line in this commit), `j` -> `stop`. It's always good to be descriptive.

pam/samplers/tour.py
```diff
@@ -305,33 +307,43 @@ def __init__(
...
+            d_zone = FrequencySampler(
+                dist=self.d_dist,
+                freq=self.d_freq,
+                threshold_matrix=self.threshold_matrix.loc[self.o_zone],
+                threshold_value=self.threshold_value,
+            ).threshold_sample()
+
+        return d_zone
 
+    def sample_destinations(self, o_loc):
```
brynpickering, line 329: type hint and docstring needed

pam/samplers/tour.py
```diff
@@ -341,9 +353,57 @@ def sequence_stops(self) -> tuple[list, list, list]:
                     "distance": ActivityDuration().model_distance(o_loc, d_facility),
                 }
             )
+        return d_seq
+
+    def create_distance_matrix(self, o_loc, d_seq):
```
brynpickering, line 358: type hints and docstrings needed

pam/samplers/tour.py
```diff
@@ -341,9 +353,57 @@ def sequence_stops(self) -> tuple[list, list, list]:
...
+        # extact o_loc coordinates into array
```
brynpickering, line 359: spelling, `extract`

pam/samplers/tour.py
```diff
@@ -341,9 +353,57 @@ def sequence_stops(self) -> tuple[list, list, list]:
...
+
+        # extract d_facility
+        d_locations = []
+        for d in [d["destination_facility"] for d in d_seq]:
+            d_locations.append([d.x, d.y])
+        d_locations = np.array(d_locations)
+
+        # Greedy TSP to minimise total travelled distance for visiting all sampled delivery locations
+        # approximated from reduce compute time
+        locs = np.concatenate([o_location, d_locations], 0)
```
brynpickering, line 370: Could simplify as:

```
locations = [[o_loc.x, o_loc.y]]
for facility in [d["destination_facility"] for d in d_seq]]:
    locations = np.append(locations, [[facility.x, facility.y]], 0)
```

pam/samplers/tour.py
```diff
@@ -341,9 +353,57 @@ def sequence_stops(self) -> tuple[list, list, list]:
...
+        seq = nx.algorithms.approximation.greedy_tsp(distance_graph, source=0)
+
+        return seq
+
+    def reorder_deliveries(self, d_seq, seq):
+        # use `seq` to re-order `d_locs` into an ordered list of dictionaries
+        # remove o_loc (first and last stop) from sequence & adjust sequence range
+        seq = seq[1:-1]
+        seq = [x - 1 for x in seq]
+        d_seq = [d_seq[order] for order in seq]
```
brynpickering, line 386:
```suggestion
        d_seq = [d_seq[order - 1] for order in seq[1:-1]]
```

brynpickering: It's shorter and I think it's a bit easier to understand what's going on

pam/samplers/tour.py
```diff
@@ -341,9 +353,57 @@ def sequence_stops(self) -> tuple[list, list, list]:
...
+
+        """
+        o_loc = self.facility_sampler.sample(self.o_zone, self.o_activity)
+
+        d_seq = TourPlanner.sample_destinations(self, o_loc)
+        dist_matrix = TourPlanner.create_distance_matrix(self, o_loc, d_seq)
+
+        seq = TourPlanner.approx_greedy_tsp(self, dist_matrix)
+
+        d_seq = TourPlanner.reorder_deliveries(self, d_seq, seq)
```
brynpickering, line 404: Call this something else so that if you come to debug at some point you have access to the original `d_seq` to compare the difference

tests/test_09_samplers_tour.py
```diff
@@ -323,6 +333,36 @@ def od_density(trips, zones_gdf, depot_density, delivery_density):
...
+
+    # test for duplicate delivery locations in d_seq
+    deliveries = set()
+    for d in d_seq:
+        d_point = d["destination_facility"]
+        assert d_point not in deliveries, "Duplicate point found in d_seq"
+        deliveries.add(d_point)
+
+    # test if o_loc is in d_seq
+    assert o_loc not in deliveries, "o_loc has been sampled as a delivery location"
```
brynpickering, line 347: separate into its own test

brynpickering (changes requested):
Some minor comments. For docstrings, the easiest approach is to set the type hints in your function signature and then use [autodocstring](https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring) in vscode to pre-generate the docstring template. Then you just need to add context.

tests/test_09_samplers_tour.py
```diff
@@ -303,6 +303,16 @@ def agent_plan(hour_sampler, minute_sampler, delivery_density, df_od, facility_s
...
+@pytest.fixture
+def d_facility_sampling(agent_plan):
+    agent_plan_test = agent_plan
+    o_loc = agent_plan_test.facility_sampler.sample(
+        agent_plan_test.o_zone, agent_plan_test.o_activity
+    )
+    d_seq = tour.TourPlanner.sample_destinations(agent_plan_test, o_loc)
+    return o_loc, d_seq
+
+
```
fredshone, line 315: fixtures make a bunch of features available (like scoping). Most of which we don't use - but I think it is still a good idea to use the fixture decorator simply to communicate to the reader that something is a fixture - so nice job.

I do find it weird that this fixture has a random sampler in it (TourPlanner) so can return different things each time it is run (?). Much prefer this return a hardcoded o_loc and d_seq.

As an aside - there is a place for randomness in testing, but not often and normally i would expect it to look like:

```
def test_sampler_bias(sampler):
   samples = [sampler.sample() for _ in range(1000)]
   assert samples.mean() < 0.5  # some statistical test
```

pam/samplers/tour.py
fredshone: avoid freight specific language

fredshone (changes requested):
NIce job with this PR.

Definately tackle Bryns comments. Getting the type hints is especially important for this module - there is loads of passing complex data types (dataframes and geodataframes) and making assumptions about column names and indexes. Tackling all these is beyond the scope of this PR but I wanted to grumble about it.

While you are here please also:

(i) Please rename ActivityDuration to something without the word "Activity" in. I found this comfusing as it is also used for trip durations.

(ii) please rename TourPlanner to something that describes the methodology (TSPPlanner maybe).

Theodore-Chatziioannou (changes requested):
That's great @val-ismaili ! I also checked the application of the method and seems to work as intended. Some minor comments only. I understand @divyasharma-arup encountered a small error when sampling with a threshold value - worth checking with her if we need to address or handle it.

pam/samplers/tour.py
divyasharma-arup: yes, I believe `d_dist` will consistently be a dataframe, and the type hint should reflect this. I thnk the confusion came from the `FrequencySampler` class, where dist can be either an Iterable or dataframe. When specifically sampling the destination distribution (`d_dist`), however, the intention is for this to be a dataframe. @val-ismaili, would you be able to update the type hint in `TourPlanner` to only be `pd.DataFrame`?

divyasharma-arup:
> That's great @val-ismaili ! I also checked the application of the method and seems to work as intended. Some minor comments only. I understand @divyasharma-arup encountered a small error when sampling with a threshold value - worth checking with her if we need to address or handle it.

Thanks Theo, I'll modify the issue with the threshold value in a separate PR -- I also need to make some changes to the ValidateTourOD class. I'm partially considering removing the threshold functionality altogether, and leaving it to the user to reduce their matrix size vs. the code doing it for them. That would require changing existing notebooks & scripts though.

pam/samplers/tour.py
val-ismaili: typehint changed to `pd.DataFrame` only

pam/samplers/tour.py
val-ismaili: language changed to refer to destinations and stops rather than deliveries.

pam/samplers/tour.py
val-ismaili: `j` -> `stop`

pam/samplers/tour.py
val-ismaili: added

pam/samplers/tour.py
val-ismaili: added

pam/samplers/tour.py
val-ismaili: agreed that can be simplified. but used an [alternative](https://github.com/arup-group/pam/pull/231/commits/51d302726821fea08161933efbd60689dd56f638#diff-47ea7e3fa3e34efd29dfa0a0980e34f9e7602e8c29f472fc4389126b8397fee9R380) way which (i think) is more efficient

pam/samplers/tour.py
val-ismaili: agreed.

pam/samplers/tour.py
val-ismaili: changed to `d_optimised_seq`

tests/test_09_samplers_tour.py
val-ismaili: seperated into additional [test](https://github.com/arup-group/pam/pull/231/commits/51d302726821fea08161933efbd60689dd56f638#diff-c548d559d784b9b8489f5833d2d6fddc2bf2b7a5244f5be4e6f9ca9b9cd69ee9R347)

pam/samplers/tour.py
brynpickering: Yep, more time efficient. Nice. Can you change one of the `d`s in your list comprehension to avoid confusion?

tests/test_09_samplers_tour.py
val-ismaili: @fredshone noted about the general point about randomness in testing. 

I purposely did it this way though because `TourPlanner.sample_destinations()` is what prevents duplicate sampling of destinations within a tour, and  the origin being sampled as a destination. By doing it this way, if someone makes a change down the line that removes this, the tests i added `test_unique_stops` and `test_origin_not_in_stops` can fail. If we hardcode `o_loc` and `d_seq` we won't be testing this and then if the `greedy_tsp()` returns an error of incomplete graph it's harder to understand where the cause is.

val-ismaili:
To summarise changes I've made following all the comments:
- Added / improved typehints & docstrings.
- General readability and simplification updates.
- I've kept the randomness in the test fixture as per this [response](https://github.com/arup-group/pam/pull/231#discussion_r1283287006).
- `ActivityDuration` class -> `DurationEstimator` (only references within `TourPlanner` class currently so easy change).
- `TourPlanner` remains the same name with added detail in the docstring describing use of TSP. Rationale is that this class may continue to develop and incorporate choices between different sequencing algorithms so keeping it generic makes sense. Also would break old scripts.

Theodore-Chatziioannou (approved):

PR #233, opened 2023-07-31
Add nbqa pre-commit hook

brynpickering:
Adding to the pre-commit hooks to clean up example notebooks following the same standards as in the python files.

Controlling line width is useful for the examples in the docs (e.g. [here](https://arup-group.github.io/pam/develop/examples/02_Population_Basics/)) as it mitigates the need to horizontally scroll (although probably doesn't remove it for most screen sizes).

## Checklist

- [ ] CHANGELOG updated
- [x] Documentation updated


fredshone (approved):
nice idea

PR #237, opened 2023-08-14
Return origin

divyasharma-arup:
Fixes # `TourPlanner` now ensures all agents modify plans to return to origin. We had an issue where agents whose tour began close to the end of the day would not return to the origin before `END_OF_DAY`. The code specifically fixes the logic in `TourPlanner.apply()`.

A test has been added to ensure this behaviour is corrected: `test_end_of_day_agent_return_depot_by_end_of_day`. To enable this test, the input test data had to be modified to allow for longer distances (and therefore a duration longer than a minute) to more easily support a test of an agent with a plan near end of day. 

Some additional tests were added to increase coverage. 

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


pam/samplers/tour.py
```diff
@@ -537,63 +521,15 @@ def add_tour_leg(
...
-            k=k,
-            zone=self.o_zone,
-            loc=o_loc,
-            activity_type="return_origin",
-            time_params=time_params,
-        )
-
-        return end_tm
-
-    def apply(self, agent: str, o_loc: Point, d_zones: list, d_locs: list) -> None:
```
divyasharma-arup, line 586: The new logic in `apply()` now makes this code redundant.

tests/test_09_samplers_tour.py
```diff
@@ -145,6 +143,23 @@ def test_model_journey_time_value():
...
+    assert sum(dist.demand.values()) == 3
+
+
+def test_distribution_total_not_zero():
+    bins = range(0, 3)
+    pivots = {0: 1, 2: 1}
+    total = 30
+    dist = tour.PivotDistributionSampler(bins=bins, pivots=pivots, total=total)
+
+    assert sum(dist.demand.values()) == 30
```
divyasharma-arup, line 160: additional tests to support coverage

fredshone (commented):
What happens in the xtreme cases? For example;

(i) a departure from depot starts at 11.50pm and requests a 30 min trip to depot?
(ii) same but for the return trip.

Is the plan completely discarded or do we get empty plans?

divyasharma-arup:
> What happens in the xtreme cases? For example;
> 
> (i) a departure from depot starts at 11.50pm and requests a 30 min trip to depot? (ii) same but for the return trip.
> 
> Is the plan completely discarded or do we get empty plans?

Hi Fred, originally, we would just have the agent stay at the origin if they did not have enough time to return to the origin. My proposed solution (for now) is to have them incrementally leave an hour earlier until they can make at least one trip. 

You also suggested 'wrapping plans' to avoid this process of checking the plans beforehand. For certain populations (HGVs), I think wrapping plans could make sense, is there an existing example of this in PAM? For LGVs, however, it might make more sense to ensure they have plans that have them at the origin by end of day. 

The problem with this approach of essentially editing the sampled plans is that the synthesised population may have what appear to be arbitrary deviations from the input samplers.

pam/samplers/tour.py
```diff
@@ -442,6 +445,60 @@ def sequence_stops(self) -> tuple[list, list, list]:
...
+            else:
+                previous_stop = d_locs[k - 1]
+                # get time from previous stop to next stop
+            stop_duration, start_tm, end_tm = DurationEstimator().model_activity_duration(
+                previous_stop, d_locs[k], end_tm
+            )
+
+            origin_duration, start_origin, end_origin = DurationEstimator().model_activity_duration(
+                d_locs[k], o_loc, end_tm
+            )
```
fredshone, line 485: I find DurationEstimator().model_activity_duration() a little confusing. Is it possible to clarify or separate activity duration and trip duration? eg:
```
time = start_tm
all_stops = [o_loc] + d_locs
for (from, to) in zip(all_stops, d_locs):
    trip_duration = DurationEstimator.trip(from, to)
    activity_duration = DurationEstimator.simple_activity(trip_duration, mini, maxi...)
    return_trip_duration = DurationEstimator.trip(o_loc, to)
    trip_end = time + trip_duration
    next_trip_start = trip_end + activity_duration
    return_time = next_trip_start + return_trip_duration
   if time_after_next_stop <= 1440:
       # etc
    # etc

    time = next_trip_start
```

pam/samplers/tour.py
```diff
@@ -442,6 +445,60 @@ def sequence_stops(self) -> tuple[list, list, list]:
...
+
+        # calculate end_tm after each stop
+        for k in range(len(d_locs)):
+            if k == 0:
+                previous_stop = o_loc
+                # get time from origin to first stop
+            else:
+                previous_stop = d_locs[k - 1]
+                # get time from previous stop to next stop
+            stop_duration, start_tm, end_tm = DurationEstimator().model_activity_duration(
```
fredshone, line 479: I think that you can more simply (and probably faster) use `DurationEstimator.model_activity_duration(...)`. ie remove paranthesis from after `DurationEstimator`. This is then called a class method (rather than object or instance method (~). ie the function belongs to `class DurationEstimator` rather than a specific instance of the `DurationEstimator`. This is possible because `DurationEstimator` doesn't have any "state" or "variables".

fredshone (changes requested):
I have a few comments. The only one that really needs addressing is the case where self.hour might become less than 0. I really appreciate that you've implemented something recursive though.

More generally agree that there is a risk that required distributions get modified by such "realism" mechanisms. What i suggest could result in shorter trips being sampled for later in the day. Which i quite like if we consider that these are agents who want to get home as you say.

But i imagine we could get more creative and insist that (i) "worker vehicles" follow more regular 9-5 schedules, (ii) "home deliveries" 8 hour shifts and sociable delivery hours (8pm?), (iii) "long haul" wrap around to 24 hours and so on.

What you've implemented is pretty close to (ii), so nice. I think the other would be unique tour planners.

re wrapping, I don't think we have anything particularly special implemented but i can show you how. It causes another weird edge case that we'd need to deal with - plans that catch up with themselves... :)

Theodore-Chatziioannou (changes requested):
I am not sure about the need to return these late-starting agents to the origin... If a freight vehicle starts late in the day, can we  not simply crop their plan at the end of the way instead of giving them a very short working day? Ie we would assume that their trip continues to the next day (not simulated), instead of having a requirement that every freight vehicle must be at the depot by midnight

divyasharma-arup:
Thanks all for the feedback, below is a summary of the problem, proposed solution for now, and changes in this branch:

**Problem:** The `TourPlanner` will return agents to origin, but this can possibly go beyond a single day. As agents add more stops (such as those with many stops), we are likely to find a fewer percentage of agents returning to their origin by end of day. This ranges from 15-25% not returning to origin, possibly higher if agents are given very long tours. The PAM `fix_plans()` method was only fixing the very last activity/leg, therefore the `TourPlanner` had to ensure only the last activity/leg was beyond a day. This has now been fixed per: [https://github.com/arup-group/pam/issues/241](url) , and this logic of breaking an agents plan past a day is no longer required in `TourPlanner`. 

**Proposed Solution:** We've discussed a few options here to address agents not returning to origin: ensuring each agent has a plan where they can return to the origin by end of day, 'wrapping' plans that go beyond a day (although we need to ensure plans cannot be more than 2 days), and introducing new specific `TourPlanners`. I'm proposing we go with the last one: a different type of TourPlanner that is based on a subpopulation's specific constraints. For example, a `TourPlanner` that has start/end/stops constraints can build a plan that meets these criteria. By distinguishing between types of agents that do tours (i.e., Service vs. Delivery tours), a `TourPlanner` that has plan constraints (i.e, must leave & return by a certain time, must do 18 stops) may be a more appropriate way to build plans vs. refining the inputs into the current `TourPlanner` to prevent multiple-day plans. For now, it's recommended that we do not have agents with a start time at the last hour of the day, as they are unlikely to have any additional legs/activities once the plan has been cropped by PAM. 

**Changes in this PR:** Mainly a reduction in code: 

- we've removed a separate `add_return_origin` function to building plans, this is now part of the `add_tour_activity` function. 
- Removed `DurationEstimator()` object, it is now a set of functions that are applied to more explicitly indicate what time is being calculated (either for an activity or leg). 
- Removed a finalise_stops method that would force agents to return to the origin by end of day. The PAM method `fix_plans()` will now crop multiple activities/legs beyond a day, and we will simply assume a second agent will be responsible for early morning tours -- effectively splitting a multi-day plan into two agents. 

If the above is satisfactory, we can merge this branch with its refactored code. I'll advocate for creating this new `TourPlanner` with input constraints as a way to add new subpopulations for a future project.

divyasharma-arup:
Let me know of any further feedback @fredshone, @Theodore-Chatziioannou

pam/samplers/tour.py
```diff
@@ -420,11 +386,11 @@ def reorder_destinations(self, d_seq, seq) -> List[Dict[str, Any]]:
 
         return d_seq
 
-    def sequence_stops(self) -> tuple[list, list, list]:
+    def sequence_stops(self) -> tuple[Point, list, list]:
```
Theodore-Chatziioannou, line 389: if you want, you could indicate that they are list of points: `list[Point]`

pam/samplers/tour.py
```diff
@@ -537,34 +501,77 @@ def add_tour_leg(
...
         end_tm = end_tm + int(trip_duration / 60)
 
         end_tm = self.add_tour_leg(
             agent=agent,
-            k=k,
-            o_zone=d_zone,
-            o_loc=d_loc,
+            k=k + 1,
+            o_zone=d_zones[len(d_locs) - 1],
+            o_loc=d_locs[len(d_locs) - 1],
```
Theodore-Chatziioannou, line 574: is this the same as `d_zones[-1]`  / `d_locs[-1]`?

Theodore-Chatziioannou (approved):
Looks good Divya! It makes sense to wrap plans instead of forcing return withing the day. Some minor styling comments, but feel free to merge.

Some ideas for the future:
- wrap overflowing plans to the next morning: ie if a plan ends at 03:00, move the 00:00-03:00 part to the morning of the same day
- joint duration distributions, where the duration probabilities are influenced by the trip start time (if we can get such observed data)

Theodore-Chatziioannou (approved):
Looks good Divya! It makes sense to wrap plans instead of forcing return withing the day. Some minor styling comments, but feel free to merge.

Some ideas for the future:
- wrap overflowing plans to the next morning: ie if a plan ends at 03:00, move the 00:00-03:00 part to the morning of the same day
- joint duration distributions, where the duration probabilities are influenced by the trip start time (if we can get such observed data)

brynpickering:
@Theodore-Chatziioannou is this ready to merge (we can ignore the failing CI, that should sort itself out once this is merged into main)

Theodore-Chatziioannou:
> @Theodore-Chatziioannou is this ready to merge (we can ignore the failing CI, that should sort itself out once this is merged into main)

waiting confirmation by @divyasharma-arup

PR #243, opened 2023-09-01
Update workflows to depend on reusable ones

brynpickering:
This will beef up the CI whilst lowering the maintenance burden by relying on workflows I've written in our very own workflow cache [here](https://github.com/arup-group/actions-city-modelling-lab) (these are mostly re-implementations of the existing PAM actions, but they can also be ported to other projects).

I've split up the CI as follows:

1. tests that run on any commit. These are quick, running on ubuntu + py3.11 and they will also push to AWS. It will  also notify in the slack feed on failure, but we might not want this?
2. tests that run when opening a PR or (re)requesting a review on a PR. These are numerous, covering multiple Python versions and all three main operating systems, as well as running the memory profiling test (so users don't have to do it themselves). The memory profiling test also results in a [flamegraph](https://bloomberg.github.io/memray/flamegraph.html) report of the memory use stored as an action artifact. For an example, see [this action run](https://github.com/arup-group/pam/actions/runs/6098705160)
3. tests that run daily. These are quick and will notify the slack feed on failure.
4. documentation building. on any commit, this will run a quick test of the build. On any push to main, this will update the `develop` version of the docs (a.k.a. "latest")
5. pre-release of a conda build. This will build a conda package when pushing a new tag starting with "v", ideally as part of a pull request. This is separate from the next step so that errors in conda package building can be caught early.
6. release of a conda build and "stable" docs. This will run when creating a GitHub release and will upload the previously built conda package to Anaconda and will update the docs to add a new version ("vX.Y.Z") and to update the "stable" alias to that version.

One thing we could change is to chain jobs in the reusable workflows themselves, not in the PAM workflows. E.g., a reusable workflow that runs a test, uploads to AWS, then notifies on slack. Then in PAM we just need a couple of lines to make all this happen. At the moment, the reusable workflows really just provide base building blocks.

## Checklist

- [ ] CHANGELOG updated
- [ ] Memory profiling tests pass (see [here](https://arup-group.github.io/pam/contributing/coding/#memory-profiling) for more details)
- [x] Documentation updated


brynpickering:
I expect memory profile tests to fail due to an issue with creating the mamba env. See https://github.com/mamba-org/setup-micromamba/issues/136

brynpickering:
Hmm, there is a slight issue with the checks that show at the bottom of the PR. 

If you click the ❌ next to [a1067b1](https://github.com/arup-group/pam/pull/243/commits/a1067b1d7e3e6fa09c51cdad6be406969e94f5c6) you will see there are many more tests that run on a PR, based on the PR being opened, a review being (re)requested, or a PR going from being in draft to "ready for review". I didn't want all these tests to run on every single commit that is pushed, but just at critical points in the review process. 

However, they have been superseded by my most recent commit push, which leads to only the "minimal" CI running (only on ubuntu, no memory profiling, no coverage report). 

Options: 

1. Leave as is and have reviewers actively check the relevant [PR CI action](https://github.com/arup-group/pam/actions/workflows/pr-ci.yml).
2. Leave as is and provide guidance on how to re-trigger the PR CI (e.g., by turning the PR into a draft and then instantly clicking "ready for review").
2. Have all pushes to a branch that has a corresponding PR trigger the PR CI. This leads to more github action time, but since this is a public repo that's fine, we aren't racking up a bill. It _does_ lead to longer CI time (minimal finishes in ~3 mins, PR CI in ~13 mins), though.

mfitz (commented):
Looks good. 

I haven't checked all the boxes for updating the changelog and whatnot, because I don't think they're relevant, but I'm happy to merge this.

mfitz (approved):

brynpickering:
Thoughts on the triggering of the PR CI @mfitz? I see you re-requested yourself for review to trigger them.

mfitz:
> Thoughts on the triggering of the PR CI @mfitz? I see you re-requested yourself for review to trigger them.

I did re-request, but that's because I (think I) mistakenly added a comment when I meant to approve the PR 🤦 

My preference is:

> 3. Have all pushes to a branch that has a corresponding PR trigger the PR CI. This leads to more github action time, but since this is a public repo that's fine, we aren't racking up a bill. It does lead to longer CI time (minimal finishes in ~3 mins, PR CI in ~13 mins), though.

I think it's fine to say: "Once you've created the PR, you get the heavy, heavy CI checks, even on subsequent pushes to that branch". If that starts to become too annoying, we can always change to the lighter options later.

fredshone (approved):
nice, we can tweak later if it becomes a pain

PR #239, opened 2023-08-18
Doc update

fredshone:
Update to docs:
- minor changes to readme intro including title logo (I've added selection of other logo assets for future use)
- new and fixed examples, including some simple re-ordering and re-naming of example notebooks

Generally aiming to ease entry barrier to new users. Not really sure if it's possible/worthwhile  to sensibly review most of this. (just trust me). 

Also ended up making a fix to the "grid optimiser" which was completely broken (woops) and added a little extra assurance to the pam.optimise tests. Not that I really expect these to get much use as is - they were intended to inspire something better...

I also relaxed nbqa ruff to allow out of order imports in notebooks. For the basic notebooks i prefer to add the imports as they are used to keep code minimal.

## Checklist

Any checks which are not relevant to the PR can be pre-checked by the PR creator. 
All others should be checked by the reviewer(s). 
You can add extra checklist items here if required by the PR.

- [x] CHANGELOG updated
- [x] Memory profiling tests pass (see [here](https://arup-group.github.io/pam/contributing/coding/#memory-profiling) for more details)
- [x] Tests added to cover contribution
- [x] Documentation updated

PR #245, opened 2023-09-08
Streamline docs

brynpickering:
I've streamlined some parts of the docs so there's less to maintain.

1. moved the Python API reference generator into the hook script, so we don't need to maintain a custom MKDocs plugin. 
2. moved tiny readme into a HTML header comment in the one file the README was referring to.
3. Added auto-generation of example notebooks in the MKDocs navigation tree, so no need to remember to put them there / update refs when renaming notebooks (as was initially forgotten in #239).

## Checklist

- [ ] CHANGELOG updated
- [x] Documentation updated


docs/static/hooks.py
```diff
@@ -1,25 +1,131 @@
...
 
-def _new_file(path: Path, config: str) -> File:
+def _new_file(path: Path, config: str, src_dir: str = ".") -> File:
+    """Link a file out in the wilderness to a filename in the documentation directory hierarchy.
+
+    Args:
+        path (Path):
+            Path (relative to "src_dir") to file that you want to bring into the documentation directory.
+            In the documentation directory, this same path with apply (e.g., `[src_dir]/path/to/file.md` will become `[docs_dir]/path/to/file.md`)
+        config (str): mkdocs config dictionary.
```
fredshone, line 36: dict i recon but no biggy

fredshone (approved):
i see you remove test - i do not complain

PR #246, opened 2023-09-08
Infer dtype of df cols to not break CI with newer pandas version

brynpickering:
Fixes the issue where a new minor release of pandas leads to more strict dtype requirements (it won't try to infer dtypes automatically). This is an issue I came across in #243 and fixed by pinning pandas more strictly, but I think it's better to fix our code.

## Checklist

- [ ] CHANGELOG updated
- [ ] Memory profiling tests pass (see [here](https://arup-group.github.io/pam/contributing/coding/#memory-profiling) for more details)
- [ ] Documentation updated


brynpickering:
I'm not entirely sure _what_ has changed in pandas v2.1.0 vs v2.0.3 to have created this issue to begin with (I don't see it in their list of backward incompatible changes). Their changelog is [here](https://pandas.pydata.org/docs/whatsnew/v2.1.0.html).

fredshone (approved):
probably we are doing something a bit derpy if we are transposing a dataframe? I thought this was going to be about ensuring the dtypes in input dataframes fro the read methods. Something we could use [pandera](https://pandera.readthedocs.io/en/stable/) for. Segwayed.

mfitz (approved):

brynpickering:
Using pandera at some point would be cool @fredshone. This is a bit of a quick and dirty fix for a bigger problem. As you say, that df generation script would benefit from a serious clean-up.

PR #249, opened 2023-09-12
Update `plot_activities` with new kwargs.

brynpickering:
Sometimes plans are very complex with long activity descriptions. 
These would benefit from more finegrain control over plotting (although, in reality, additions to this control could continue indefinitely).

Here I've just added the ability to control plot width (to fit more text) and label fontsize (to allow a user to change fontsize to better fill activity rectangles).

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


tests/test_10_plotting.py
```diff
@@ -178,14 +182,31 @@ def test_plot_travel_plans_for_household(instantiate_household_with, cyclist, pt
...
+def test_plot_activities_with_label_fontsize_partial(person_df):
+    plot_activities(person_df, label_fontsize={"Home": 20})
+
+
+def test_plot_activities_with_label_fontsize_full(person_df):
+    plot_activities(person_df, label_fontsize={"Home": 20, "Education": 5, "Travel": 15})
+
+
+def test_plot_activities_with_new_width(person_df):
+    plot_activities(person_df, width=40)
```
KasiaKoz, line 207: these don't have assertions, what are they testing?

brynpickering: That the function doesn't raise an exception of some kind. It's a bit of an annoying function to test as it doesn't return anything. One option would be to return fig and axs from it and then inspect the details (e.g., get the fontsize of labels).

KasiaKoz: might be clearer to do something like this:
```
try:
    plot_activities(person_df, width=40)
except Exception as e:
    raise AssertionError(f'test failed when plot_activities raised {e.__repr__()}')
```
https://stackoverflow.com/questions/4319825/python-unittest-opposite-of-assertraises

or instead of testing for no exceptions, could check that the the function that is using the kwarg being tested is called with expected values, so for width it looks like `plt.subplots` uses width, so could do something like this maybe:

```
def test_plot_activities_with_new_width(person_df, mocker):
    mocker.patch.object(plt, 'subplots')

    plot_activities(person_df, width=40)
    
    plt.subplots.assert_called_once_with(a,b,c,d)
```

tests/test_10_plotting.py
```diff
@@ -178,17 +183,61 @@ def test_plot_travel_plans_for_household(instantiate_household_with, cyclist, pt
...
+        fc = patch.get_facecolor()
+        assert cmap[label] + (1,) == fc
+
+
+def test_plot_activities_fontcolor(person_df):
+    cmap = {"Home": (1, 1, 1), "Education": (0, 0, 0), "Travel": (0.3, 0.3, 0.3)}
+    expected_fontcolor = {"Home": "black", "Education": "white", "Travel": "white"}
+    fig, ax = plot_activities(person_df, cmap=cmap)
+    for text in ax.texts:
+        assert text.get_color() == expected_fontcolor[text.get_text()]
```
KasiaKoz, line 212: I don't understand how this works, the font colour doesn't seem to be passed to the function, so how is it making it's way there?

tests/test_10_plotting.py
```diff
@@ -178,17 +183,61 @@ def test_plot_travel_plans_for_household(instantiate_household_with, cyclist, pt
...
+
+
+def test_plot_activities_with_label_fontsize_partial(person_df):
+    fontsizes = {"Home": 20}
+    fig, ax = plot_activities(person_df, label_fontsize={"Home": 20})
+    for text in ax.texts:
+        assert text.get_fontsize() == fontsizes.get(text.get_text(), 10)
```
KasiaKoz, line 219: the 10 here is the method's default right? might be good to name it as a variable, it'll make it easier to update if that default changes in the future, but also it's just some random 10 otherwise

tests/test_10_plotting.py
```diff
@@ -178,17 +183,61 @@ def test_plot_travel_plans_for_household(instantiate_household_with, cyclist, pt
...
+
+def test_plot_activities_with_label_fontsize_full(person_df):
+    fontsizes = {"Home": 20, "Education": 5, "Travel": 15}
+    fig, ax = plot_activities(person_df, label_fontsize=fontsizes)
+    for text in ax.texts:
+        assert text.get_fontsize() == fontsizes[text.get_text()]
+
+
+def test_plot_activities_with_new_width(person_df):
+    width = 40
```
KasiaKoz, line 230: why not use this variable again below instead of `40`?

tests/test_10_plotting.py
```diff
@@ -178,17 +183,61 @@ def test_plot_travel_plans_for_household(instantiate_household_with, cyclist, pt
...
+    assert not any(isinstance(i, Legend) for i in fig.get_children())
+
+
+def test_plot_activities_with_cmap(person_df):
+    cmap = {"Home": (1, 1, 1), "Education": (0, 0, 0), "Travel": (0.3, 0.3, 0.3)}
+    fig, ax = plot_activities(person_df, cmap=cmap)
+    for idx, patch in enumerate(ax.patches):
+        label = ax.texts[idx].get_text()
+        fc = patch.get_facecolor()
+        assert cmap[label] + (1,) == fc
```
KasiaKoz, line 204: what is `+ (1,)` doing here?

KasiaKoz (commented):
Minor comments, I think in general, with the esoteric fig/ax testing going on now, it would help if the test names revealed more about what the tested/expected outcome is.

There is something else though that is less minor. I ran one of the expected notebooks (with no changes) and the plots seem a bit derpy, the legend is running away from the plots and they are a bit more squished than before. Are we expecting the plots to look the same as before if there are no kwargs passed?

Before:
Screenshot 2023-09-19 at 10 17 31
After:
Screenshot 2023-09-19 at 09 43 27

tests/test_10_plotting.py
brynpickering: The font colour is based on perceived luminance of the rectangle colour (see the plotting source code). This is a slight change from before, where "Travel" would always have a _white_ font colour and all other font colours would be _black_. This change fixes the underlying issue that if a user ever defined their own `cmap` with dark backgrounds, they would end up not being able to see the label text as it would default to _black_.

brynpickering:
OK the issue with the weird formatting should now be fixed.

Before:
image

After:
image

brynpickering:
memory profiler is failing due to a timeout (rather than exceeding memory). I tested locally and it isn't a problem. I think I might remove the timeout part of that test as the GitHub runner seems to really skew runtimes...

pam/plot/plans.py
```diff
@@ -166,24 +167,25 @@ def build_cmap(df: pd.DataFrame) -> dict:
 def plot_activities(
     df: pd.DataFrame,
     cmap: Optional[dict] = None,
-    width: int = 16,
+    width: int = DEFAULT_ACTIVITIES_PLOT_WIDTH,
```
KasiaKoz, line 170: i like these global default values

KasiaKoz (approved):

PR #252, opened 2023-09-13
location model for discretionary activities

Theodore-Chatziioannou:
This PR applies a time-prism approach for choosing the location of non-mandatory activities. With this approach, agents try to sensibly combine the location of their discretionary activities around the locations of “anchor points” (defined as the locations of mandatory activities – home, employment, education).

The pam.planner.choice_location.DiscretionaryTrips class allows the user to apply the time-space prism approach on a PAM plan, given accessibility and attraction data.

The typical workflow goes as follows:
```
model = DiscretionaryTrips(population, od) # initialize the model and point to the data objects 
model.update_plan() # apply the model and update the location of each non-mandatory activity in the plan
```

We start by identifying fixed mandatory locations (ie home, work, education), splitting the PAM plan into chains of activities.

We then distinguish between 4 possible cases of trip chains:

example-discretionary-cases

Case 1 is a simple tour between mandatory activities, and may be approached with a discrete choice model (see 15_advanced_choice_modelling.ipynb).

Case 2 selects a possible destination using origin-destination probabilities from the anchor location (saved under the od_probs variable of the od object)

Case 3 weights candidate locations as a function of three elements:
- leg ratio
- diversion factor
- attraction (OD probabilities)

Case 4 solves the first location as case 2, and then the rest as case 3.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


brynpickering:
@Theodore-Chatziioannou the code needs linting/formatting. Two options:

Install `pre-commit` in your dev environment: `pre-commit install` 

Then run it on all files: `pre-commit run --all-files`

With pre-commit installed, future commits will be auto-cleaned.

Theodore-Chatziioannou:
> pre-commit run --all-files

nice - done

tests/conftest.py
brynpickering, line 1: Additions here seem like a maintenance nightmare. I have an activity sequence generator that I've put together locally that I'll add to this branch once you've clarified some other comments.

tests/conftest.py
```diff
@@ -1152,3 +1153,889 @@ def labels():
...
+            start_time=mtdt(18 * 60),
+            end_time=mtdt(19 * 60),
+        )
+    )
+    plan.add(
+        Leg(
+            seq=3,
+            mode="car",
+            start_area="h",
+            end_area="b",
```
brynpickering, line 1323: Is "b" the correct end area? The next activity happens in area "c".

tests/conftest.py
```diff
@@ -1152,3 +1153,889 @@ def labels():
...
+            start_time=mtdt(6 * 60),
+            end_time=mtdt(17 * 60),
+        )
+    )
+    plan.add(
+        Leg(
+            seq=2,
+            mode="car",
+            start_area="w",
+            end_area="a",
```
brynpickering, line 1401: Is "a" the correct end area? The next activity happens in area "c".

tests/conftest.py
```diff
@@ -1152,3 +1153,889 @@ def labels():
...
```
brynpickering, line 1479: Is "a" the correct end area? The next activity happens in area "c".

tests/test_24_planner_choice_location.py
```diff
@@ -132,3 +155,45 @@ def test_utility_calculation(choice_model_mnl):
...
     )
+
+
+def test_match_leg_ratio_probabilities(discretionary_trip_od: DiscretionaryTripOD):
+    assert approx(discretionary_trip_od.leg_ratios, abs=0.005) == [0.667, 0.500, 0.333]
+    assert approx(discretionary_trip_od.leg_ratio_p, abs=0.005) == [0.444, 0.667, 0.889]
+
+
+def test_match_diversion_probabilities(discretionary_trip_od: DiscretionaryTripOD):
+    assert approx(discretionary_trip_od.diversion_p, abs=0.05) == [0.545, 0.545, 0.545]
```
brynpickering, line 166: Should `abs` be `0.005`?

tests/test_24_planner_choice_location.py
```diff
@@ -132,3 +155,45 @@ def test_utility_calculation(choice_model_mnl):
...
+
+
+def test_match_attraction_probabilities(discretionary_trip_od: DiscretionaryTripOD):
+    assert approx(discretionary_trip_od.attraction_p, abs=0.005) == [0.231, 0.306, 0.463]
+
+
+def test_match_destination_probabilities(discretionary_trip_od: DiscretionaryTripOD):
+    assert approx(discretionary_trip_od.destination_p, abs=0.005) == [
```
brynpickering, line 174: Should `abs` be the same order of magnitude as the numbers (which are more like `0.00000005`)?

tests/conftest.py
```diff
@@ -1152,3 +1153,889 @@ def labels():
...
+                zone_labels,
+                zone_labels,
+                np.array([[500, 333, 250], [333, 1000, 333], [250, 333, 500]]),
+            ),
+            ODMatrix(
+                "od_probs",
+                "bus",
+                zone_labels,
+                zone_labels,
+                np.array([[333, 250, 222], [250, 1000, 250], [222, 250, 333]]),
```
brynpickering, line 1203: This was something I noticed in the example notebook. Why are the probabilities > 1? Is this standard for OD matrices?

examples/17_advanced_discretionary_locations.ipynb
```diff
@@ -0,0 +1,1007 @@
...
+    "\n",
+    "from pam.activity import Activity, Leg, Plan\n",
+    "from pam.location import Location\n",
+    "from pam.planner.choice_location import DiscretionaryTripOD, DiscretionaryTrips\n",
+    "from pam.planner.od import ODFactory, ODMatrix\n",
+    "from pam.planner.utils_planner import get_trip_chains_either_anchor\n",
+    "from pam.utils import minutes_to_datetime as mtdt\n",
+    "from pam.variables import END_OF_DAY\n",
+    "\n",
+    "logging.basicConfig(level=logging.DEBUG)\n",
```
brynpickering, line 47: Can you try `logging.getLogger('pam').setLevel(logging.DEBUG)` to avoid getting aloads of matplotlib debug logs? If this doesn't work then we should definitely open an issue to create a pam-specific logger that allows users to change the log level for pam-only log messages.

examples/17_advanced_discretionary_locations.ipynb
```diff
@@ -0,0 +1,1007 @@
...
+       "array([250., 333., 500.])"
+      ]
+     },
+     "execution_count": 20,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "model._od[\"od_probs\", \"w\", :, \"car\"]"
```
brynpickering, line 863: If you are inspecting a private property in the example notebook, perhaps there should be a public interface to it:

``` python
@property
def od():
    return self._od
```

brynpickering (changes requested):
Nicely packaged implementation @Theodore-Chatziioannou ! I have some clean-up and clarification comments. I'll leave @panostsolerid to comment on the value of the implementation itself.

tests/conftest.py
```diff
@@ -1152,3 +1153,889 @@ def labels():
...
+    )
+
+    return plan
+
+
+@pytest.fixture
+def plan_home_shop_shop_work_shop_shop_home():
+    plan = Plan(home_area="h", home_location=Location(loc=Point(0, 0), area="h"))
+    plan.add(
+        Activity(
```
brynpickering, line 1904: activity ends at 5h but next leg starts at 2h

tests/conftest.py
```diff
@@ -1152,3 +1153,889 @@ def labels():
...
+            seq=1, act="home", area="h", loc=Point(0, 0), start_time=mtdt(0), end_time=mtdt(5 * 60)
+        )
+    )
+    plan.add(
+        Leg(
+            seq=1,
+            mode="car",
+            start_area="h",
+            end_area="z",
```
brynpickering, line 1913: are all shops in this sequence expected to be in "z"?

tests/conftest.py
```diff
@@ -1152,3 +1153,889 @@ def labels():
...
+            end_loc=Point(0, 0),
+            start_time=mtdt(7 * 60),
+            end_time=mtdt(8 * 60),
+        )
+    )
+    plan.add(
+        Activity(
+            seq=4,
+            act="home",
+            area="b",
```
brynpickering, line 1824: should "home" be in area "h"?

tests/conftest.py
```diff
@@ -1152,3 +1153,889 @@ def labels():
...
+            start_time=mtdt(7 * 60),
+            end_time=mtdt(8 * 60),
+        )
+    )
+    plan.add(
+        Activity(
+            seq=4,
+            act="work",
+            area="w",
+            loc=Point(0, 0),
```
brynpickering, line 1969: work is in the same location as home. By design?

pam/planner/choice_location.py
```diff
@@ -4,18 +4,22 @@
 from collections.abc import Callable
 from copy import deepcopy
 from dataclasses import dataclass
-from typing import Literal, NamedTuple, Optional, Union
+from typing import List, Literal, NamedTuple, Optional, Union
```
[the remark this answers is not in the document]
Theodore-Chatziioannou: both are valid - I used `List` to keep compatibility with v3.7. I see now that we have dropped support for 3.7 in other places, so I am renaming.

pam/planner/choice_location.py
```diff
@@ -253,3 +257,246 @@ def __init__(self, population: Population, od: OD, zones: pd.DataFrame) -> None:
...
+        b) O->discretionary->D (DiscretionaryTripOD)
+        c) O->discretionary->discretionary->O
+        d) O->discretionary->discretionary->D
+
+    Chains with multiple discretionary trips are solved
+        recursively, updating the first location each time,
+        and then keeping it fixed as we solve downstream.
+    """
+
+    def __init__(self, trip_chain: List[Union[Activity, Leg]], od: OD) -> None:
```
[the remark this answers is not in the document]
Theodore-Chatziioannou: you are right, PEP suggests keeping the docstrings under init. Moved

pam/planner/utils_planner.py
```diff
@@ -69,3 +93,68 @@ def get_validate(obj, name: str):
     if attr is None:
         raise ValueError(f"Attribute {name} has not been set yet")
     return attr
+
+
+def get_act_names(seqs: List[Union[Activity, Leg]]) -> List[str]:
```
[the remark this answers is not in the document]
Theodore-Chatziioannou: yes, this is quite generic, and can have applications in other future modules within pam.planner

pam/planner/utils_planner.py
```diff
@@ -69,3 +93,68 @@ def get_validate(obj, name: str):
...
+    elif chain[-1].act not in LONG_TERM_ACTIVITIES and chain[0].act in LONG_TERM_ACTIVITIES:
+        leg = deepcopy(chain[1])
+        act = deepcopy(chain[0])
+        chain.append(leg)
+        chain.append(act)
+    # if all activities are non-disrectionary, create a tour from/to the first location
+    elif chain[-1].act not in LONG_TERM_ACTIVITIES and chain[0].act not in LONG_TERM_ACTIVITIES:
+        act = deepcopy(chain[0])
+        leg = deepcopy(chain[1])
+        act.act = "home"
```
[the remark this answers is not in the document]
Theodore-Chatziioannou: this step adds an "anchor" to a trip chain that does not have one. Here, we make the simple assumption that a chain starts at home, which is at the same zone as the first activity. This is an edge case really, typically plans should start with a "home" activity.

tests/conftest.py
Theodore-Chatziioannou: updated

tests/conftest.py
Theodore-Chatziioannou: yep. The idea is that the "discretionary planner" will allocate them to a different zone.

tests/conftest.py
Theodore-Chatziioannou: a generator sounds good! Indeed the current process is a bit unwieldy, with a lot of data repetition and scope for errors.

pam/planner/choice_location.py
```diff
@@ -253,3 +257,246 @@ def __init__(self, population: Population, od: OD, zones: pd.DataFrame) -> None:
...
+      1) Distribution of leg compared to total trip
+      2) Diversion factor (compared to direct trips)
+      3) Zone attraction
+    Final probabilities are defined as (1) * (2) * (3)
+      (and then normalised to sum up to 1)
+
+    """
+
+    @staticmethod
+    def pdf_leg_ratio(leg_ratios: np.array, observed_ratio: float) -> np.array:
```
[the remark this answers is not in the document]
Theodore-Chatziioannou: technically yes, but you are not really expected to provide a list of lists here. Simple array communicates intention better here I believe

pam/planner/choice_location.py
```diff
@@ -253,3 +257,246 @@ def __init__(self, population: Population, od: OD, zones: pd.DataFrame) -> None:
...
+
+        # anchor points
+        self.anchor_zone_start = trip_chain[0].location.area
+        self.anchor_zone_end = trip_chain[-1].location.area
+        self.trmode = trip_chain[1].mode
+
+        # self.anchor_zone_start = self._od.encode(self.anchor_area_start)
+        # self.anchor_zone_end = self._od.encode(self.anchor_area_end)
+
+    def choose_destination(self):
```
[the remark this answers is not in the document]
Theodore-Chatziioannou: good idea - updated

examples/17_advanced_discretionary_locations.ipynb
Theodore-Chatziioannou: updated

pam/planner/choice_location.py
```diff
@@ -253,3 +257,246 @@ def __init__(self, population: Population, od: OD, zones: pd.DataFrame) -> None:
...
+        and then keeping it fixed as we solve downstream.
+    """
+
+    def __init__(self, trip_chain: List[Union[Activity, Leg]], od: OD) -> None:
+        self._trip_chain = trip_chain
+        self._od = od
+        self.act_names = get_act_names(trip_chain)
+
+        # anchor points
+        self.anchor_zone_start = trip_chain[0].location.area
```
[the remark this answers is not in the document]
Theodore-Chatziioannou: added assertions to the trip chain functions

pam/planner/choice_location.py
```diff
@@ -253,3 +257,246 @@ def __init__(self, population: Population, od: OD, zones: pd.DataFrame) -> None:
...
+        Update the PAM activity locations of the first activity activity
+            in the trip chain,
+            and continue downstream until the entire chain is solved.
+
+        Returns:
+            None
+
+        """
+        if len(self.act_names) > 2:
+            assert isinstance(self._trip_chain[0], Activity)
```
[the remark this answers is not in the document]
Theodore-Chatziioannou: moved to initialisation.

examples/17_advanced_discretionary_locations.ipynb
Theodore-Chatziioannou: thanks, that worked!

Theodore-Chatziioannou:
thanks @brynpickering , I have updated the PR in light of your comments. The activity generator sounds like a good addition once the PR is merged. Let me know if anything else.

brynpickering (approved):

PR #253, opened 2023-09-19
Ipf

Theodore-Chatziioannou:
Introduces methods for creating a simple synthetic population using an Iterative Proportional Fitting (IPF) approach.

A resampled population can be generated by providing a seed PAM population, and a dataframe with zone marginals.

```
from pam.planner import ipf
population = ipf.generate_population(seed_population, df_zones)
```

Example provided under `./examples/18_advanced.ipynb`.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Python versions 3.9 to 3.11 support ([#192], [#210]).
 - Documentation, now available at https://arup-group.github.io/pam ([#197]).
 - Time-space prism method for selecting the location of non-mandatory activities ([#252](https://github.com/arup-group/pam/pull/252)).
+- Simple IPF approach for generating synthetic populations ([#253]).
 - **internal** [Codecov](https://codecov.io) and [pre-commit](https://pre-commit.ci/) CI bots ([#202]).
 - **internal** Github action job to build PAM and run tests on a Windows machine ([#192]).
 - **internal** Contribution guidelines and issue/pull request templates ([#207]).
--- a/pam/planner/ipf.py
+++ b/pam/planner/ipf.py
@@ -87,15 +87,18 @@ def prepare_zone_marginals(zone_data: pd.DataFrame) -> tuple[dict, dict[str, lis
     Args:
         zone_data (pd.DataFrame): Zone data, with the zone label as the dataframe index.
             The dataframe columns should follow this convention: `variable|class`,
-            for example: `age|minor, age|adult, income|low, income|high, ....`
+            for example: `age|minor, age|adult, income|low, income|high, ....`/
+            Alternatively, the user can provide a multi-index column format,
+            where the first level represents the variable, and the second the class.
 
     Returns:
         tuple[dict, dict[list[np.array]]]: Zone encodings and marginals.
     """
     df_marginals = zone_data.copy()
-    df_marginals.columns = pd.MultiIndex.from_tuples(
-        [tuple(x.split("|")) for x in df_marginals.columns]
-    )
+    if df_marginals.columns.nlevels != 2:
+        df_marginals.columns = pd.MultiIndex.from_tuples(
+            [tuple(x.split("|")) for x in df_marginals.columns]
+        )
 
     encodings = defaultdict(list)
     for x, y in df_marginals.columns:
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -49,6 +50,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Example data files not accessed by any example notebook ([#196]).
 - **internal** Unused scripts that were outside the source code directory ([#199]).
 
+## [v0.2.5] - 2023-09-13
```
brynpickering: remove / move this into unreleased

pam/planner/ipf.py
```diff
@@ -0,0 +1,220 @@
...
+
+from pam.core import Person, Population
+
+
+def safe_divide(x: np.array, y: np.array) -> np.array:
+    """Safely divide two arrays. When dividing by zero, the result is set to zero.
+
+
+    Args:
+        x (np.array): Nominator array.
```
brynpickering, line 18:
```suggestion
        x (np.array): Numerator array.
```

pam/planner/ipf.py
```diff
@@ -0,0 +1,220 @@
...
+from copy import deepcopy
+from typing import Optional
+
+import numpy as np
+import pandas as pd
+
+from pam.core import Person, Population
+
+
+def safe_divide(x: np.array, y: np.array) -> np.array:
```
brynpickering, line 13: This feels like a util function that should be in a utils file

brynpickering (commented):
Nothing jumps out at me as problematic here, expect perhaps the need for pipes in the dataframe columns. I would just require multiindex columns directly.

One question I do have though is how well you feel this performs in terms of time/memory. Do you have a feel for whether your implementation is sufficiently performant at scale? Nested loops always give rise to concern that it might not be ;)

pam/planner/ipf.py
```diff
@@ -0,0 +1,220 @@
...
+    Args:
+        zone_data (pd.DataFrame): Zone data, with the zone label as the dataframe index.
+            The dataframe columns should follow this convention: `variable|class`,
+            for example: `age|minor, age|adult, income|low, income|high, ....`
+
+    Returns:
+        tuple[dict, dict[list[np.array]]]: Zone encodings and marginals.
+    """
+    df_marginals = zone_data.copy()
+    df_marginals.columns = pd.MultiIndex.from_tuples(
```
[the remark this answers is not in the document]
Theodore-Chatziioannou: I know, it's not ideal, but I wanted something that a) can be easily read from the disk, and b) follows the usual format of the "zones data" table used elsewhere in the planner module.

pam/planner/ipf.py
Theodore-Chatziioannou: moved

pam/planner/ipf.py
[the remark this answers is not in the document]
brynpickering: I guess it depends on who you imagine your user to be. If you don't think they should have any ability to use Pandas, fair enough. But I suspect they should have some Pandas knowledge, in which case Multiindex data is easy to read from disc, e.g.,

image

can be read with `pd.read_csv("path/to/data.csv", index_col=0, header=[0, 1])`.
This has the advantage of being easier to prepare "by hand" if needs be and is easier to slice in excel/pandas before applying the IPF treatment.

Then defining directly in a notebook becomes one of:

``` python
zone_data = pd.DataFrame(
   {
        ("hhincome", "high"): {"a": 30, "b": 80},
        ("hhincome", "medium"): {"a": 40, "b": 100},
        ("hhincome", "low"): {"a": 30, "b": 20},
        ("age", "minor"): {"a": 40, "b": 90},
        ("age", "adult"): {"a": 60, "b": 110},
        ("carAvail", "yes"): {"a": 90, "b": 180},
        ("carAvail", "no"): {"a": 10, "b": 20},
    },
).rename_axis(index="zone")
```

or 

``` python
zone_data = pd.DataFrame(
   {
        ("hhincome", "high"): [30, 80],
        ("hhincome", "medium"): [40, 100],
        ("hhincome", "low"): [30, 20],
        ("age", "minor"): [40, 90],
        ("age", "adult"): [60, 110],
        ("carAvail", "yes"): [90, 180],
        ("carAvail", "no"): [10, 20],
    }, index=pd.Index(data=["a", "b"], name="zone")
)
```

I'm not going to fight too hard for this, but it does seem easier to place the burden on the user than on PAM to format the dataframe correctly here.

pam/planner/ipf.py
```diff
@@ -0,0 +1,222 @@
...
+    """
+    person_pool = defaultdict(list)
+    for hid, pid, person in population.people():
+        code = tuple([v.index(person.attributes[k]) for k, v in encodings.items()])
+        person_pool[code].append(person)
+
+    return person_pool
+
+
+def sample_population(encodings, dist, sample_pool) -> Population:
```
brynpickering, line 161: Missing type hints and Arg defs in the docstring (I think this is the only one!)

pam/planner/ipf.py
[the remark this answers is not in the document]
Theodore-Chatziioannou: I have made it an option (the user can now use either format)

pam/planner/ipf.py
Theodore-Chatziioannou: updated

pam/planner/ipf.py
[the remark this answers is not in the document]
brynpickering: OK, cool. Could you add examples of both options to the example notebook?

brynpickering (approved):

PR #254, opened 2023-09-20
Remove timing benchmark

brynpickering:
We're seeing odd behaviour with the timing benchmark in the memory profiling. I had already moved it from 280 seconds to 400 seconds to handle the GitHub Action runners seemingly being slower than local machines. However, in e.g. #249, this 400 second limit was breached (even though locally it still runs in under 280 secs).

The memory profiling is mainly meant to check that memory limits aren't breached, so I think it's OK to remove the timing benchmark. We can always see how long the CI test take in total to get an order-of-magnitude understanding of whether some central component in PAM now leads to long runtimes.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated

### DIFF
--- a/tests/test_100_memory_profiling.py
+++ b/tests/test_100_memory_profiling.py
@@ -6,13 +6,11 @@ import pytest
 from pam import read
 
 BENCHMARK_MEM = "2890.59 MB"
-BENCHMARK_SECONDS = 400
 
 data_dir = Path(__file__).parent / "test_data"
 
 
 @pytest.mark.limit_memory(BENCHMARK_MEM)
-@pytest.mark.timeout(BENCHMARK_SECONDS)
 @pytest.mark.high_mem
 def test_activity_loader():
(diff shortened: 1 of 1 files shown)


Theodore-Chatziioannou (approved):
maybe this test module could only run periodically, or on-demand, or after specific changes.. It is testing a single function with only few dependencies, which will not change often. So if it is taking more than 5 minutes to run for every single commit, it may be an overkill..

brynpickering:
> maybe this test module could only run periodically, or on-demand, or after specific changes.. It is testing a single function with only few dependencies, which will not change often. So if it is taking more than 5 minutes to run for every single commit, it may be an overkill..

We discussed this in #243. We decided to have it run on every push to a PR. It basically means that once you decide your branch is PR-ready, it doesn't really matter if we have this test that takes a bit longer as it shouldn't need to run too often and can run while the reviewer is checking the code.

Issue #255, opened 2023-09-22
Abstractmethod docstrings are not inherited in API reference


brynpickering:
### Description

As is being tracked [here](https://github.com/mkdocstrings/griffe/issues/96), we do not get docstrings in our reference API for abstract methods that are set by child classes, i.e.,

```python
class DiscretionaryTrip(ABC):
    ...
    @abstractmethod
    def choose_destination(self) -> str:
        """Selects a destination for the discretionary activity.

        Returns:
            str: Selected destination zone name.
        """

class DiscretionaryTripOD(DiscretionaryTrip):
    ...
    def choose_destination(self) -> str:
        zone = sample_weighted(self.destination_p)
        area = self._od.labels.destination_zones[zone]

        return area
```

Python passes the abstract method docstring down to the child class (`help(my_discretionary_trip_od_model.choose_destination)` will return the expected docstring), but this isn't the case in the docs:

![image](https://github.com/arup-group/pam/assets/17178478/783b02b1-7eb2-4e09-a563-6b631fb246d8)

### Related links

- https://arup-group.github.io/pam/develop/reference/pam/planner/choice_location/#pam.planner.choice_location.DiscretionaryTripOD.choose_destination

### Version

0.2

### Proposed change

Repeat docstring in child class methods until https://github.com/mkdocstrings/griffe/issues/96 is fixed.

PR #256, opened 2023-09-22
Clean docstrings and refs

brynpickering:
A few things changed here:

1. A temporary fix for #255 (which should stay open after merging this PR since we are ultimately waiting for an upstream fix) by reintroduing duplicate docstrings in child classes of `DiscretionaryTrip`.
2. Removed all un-referenced graphics. They can always be resurected from old commits if a user wants to reference them again.
3. Better cross-referencing in the docs, particlarly to the PAM API docs (e.g., [][pam.core.Person] will create a link to https://arup-group.github.io/pam/develop/reference/pam/core/#pam.core.Person).
4. Add member inheritance, so that child classes show their parent class methods in the API docs too. It adds duplication to the API docs but does make it easier to view what methods are available for a class without needing to follow the inheritance chain.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated


pam/planner/choice_location.py
```diff
@@ -446,8 +439,7 @@ def leg_ratio_p(self) -> np.array:
     @property
     def diversion_factors(self) -> np.array:
         """
```
Theodore-Chatziioannou, line 441: why are you re-introducing long lines here?

Theodore-Chatziioannou (approved):
looks good, but I would rather revert the `set` change to `pd.unique` before merging, as per my comment above.

pam/planner/od.py
```diff
@@ -107,10 +106,10 @@ def from_matrices(cls, matrices: list[ODMatrix]) -> OD:
     @staticmethod
     def prepare_labels(matrices: list[ODMatrix]) -> Labels:
         labels = Labels(
-            vars=list(pd.unique([mat.var for mat in matrices])),
+            vars=list(dict.fromkeys(mat.var for mat in matrices)),
```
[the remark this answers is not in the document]
brynpickering: I've updated to a method that preserves order. This fixes a deprecation warning from pandas since they will not allow you to apply `pd.unique` to non-pandas objects in a future version

pam/planner/choice_location.py
brynpickering: docstrings wrap in tooltips etc. anyway, so it is better to write them more "naturally" and let the downstream tools to decide where the linebreaks are. 

I tend to opt to set linebreaks at the end of sentences. This also helps to ensure you're not accidentally writing excessively long sentences (as you can visually parse the length of your sentences quickly). This is good because long sentences will lead to poor information transfer!

There's obviously a compromise with legibility in the code itself (lines that are too long might be difficult to parse). This is partially solved by keeping sentences short. But also most IDEs allow line wrapping which will automatically deal with long docstring lines.

PR #258, opened 2023-09-25
Fix long run time and reintroduce timing benchmark test

brynpickering:
Fixes long memory profiling CI runtimes.

It seems that the timeout that I "fixed" by removing in #254 was actually trying to tell us something. A change in how pandas groupby works means that the current implementation on reading data became _unbelievably_ slow. To be honest, I'm surprised that pivoting a dataframe into a nested dict was applied for "performance" reasons to begin with... 

Anyway, I've got rid of the df->dict conversion. Locally, this not only makes it possible to run the timing/memory benchmark (CI was taking __hours__ for that test!) but actually halves the time it takes to run! Will be interesting to see if this leads to real speed-ups in PAM in more practical applications.

EDIT: this update leads to slower runtimes than the previous method after @Theodore-Chatziioannou tested it with a real dataset (10-20% slower). We should fix this and the profiling dataset so that it is catching runtimes that align better with real PAM usecases.

UPDATE: memory requirements also half when making this change. The benchmark is now ~1.4GB instead of ~2.8GB.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - optimise.grid.grid_search fixed ([#239]).
 - `TourPlanner` prevents sampling of duplicate destinations, and prevents origin being sampled as a destination ([#231]).
 - Fix for [#221](https://github.com/arup-group/pam/issues/221), improved "pt simplification" ([#222])
+- Slow loading of data with e.g., [pam.read.load_travel_diary][pam.read.diary.load_travel_diary] when using pandas v2.1.1 (caused by `pandas.MultiIndex.groupby`, see [pandas issue #55256](https://github.com/pandas-dev/pandas/issues/55256)). ([#258])
 
 ### Added
 - MATSim warm starting example ([#239]).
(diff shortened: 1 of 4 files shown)


fredshone:
I think Theo added this df -> dict step as a speed up N years ago. Curious to know what has happened to groupby to spoil this.

Theodore-Chatziioannou:
yes, let me check with a large population... Our finding back then was the opposite: that `.loc` was extremely slow, and using a dictionary instead gave us a performance improvement.. But this may not be the case anymore..

brynpickering:
`.loc` being slow seems to not be the case as of pandas v1.5.0 (see https://github.com/pandas-dev/pandas/issues/23735). So it may well have been true for an old version of PAM (although the memory overhead introduced by df -> dict probably existed either way). Anyway, with new pandas the pendulum has swung the other way, so seems worth taking advantage of the benefits.

brynpickering:
RE why groupby is now so slow, it could be a regression introduced by the latest version of pandas (e.g., https://github.com/pandas-dev/pandas/issues/52070), but I kinda doubt it since we're not actually operating on the data at all, just adding grouped arrays to a dictionary...

brynpickering:
Looking at where it gets stuck on `df` -> `dict` with pandas v2.1.1, this seems to be the associated issue: https://github.com/pandas-dev/pandas/issues/55256

Theodore-Chatziioannou (approved):
tested against a large pop: loading was slightly slower to before, but not a large difference. Suggest merging and revising the indexing approach in the future.

PR #264, opened 2023-10-30
fix bug

Theodore-Chatziioannou:
Fix for #263 .

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/tests/test_03_read_matsim.py
+++ b/tests/test_03_read_matsim.py
@@ -220,4 +220,4 @@ def test_get_float_attribute_from_person():
     </person>"""
     elem = et.fromstring(text)
     pid, attributes = get_attributes_from_person(elem)
-    assert isinstance(attributes['age'], float)
+    assert isinstance(attributes["age"], float)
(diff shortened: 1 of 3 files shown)


tests/test_03_read_matsim.py
```diff
@@ -210,3 +210,14 @@ def test_get_attributes_from_person():
     assert pid == "chris"
     assert attributes["hid"] == "A"
     assert attributes["vehicles"] == {"car": "chris"}
+
+
+def test_get_float_attribute_from_person():
```
mfitz, line 215: When you ran this test before fixing the bug, did it throw an exception?

mfitz (approved):

tests/test_03_read_matsim.py
Theodore-Chatziioannou: yep

Theodore-Chatziioannou: it was tring to cast a lxml element as float (`float(attr)`), instead of the element value (`attr.text`)

brynpickering (approved):
Good catch!

brynpickering:
@Theodore-Chatziioannou you need to run pre-commit locally!

Theodore-Chatziioannou:
> @Theodore-Chatziioannou you need to run pre-commit locally!

`pre-commit run` ? It is not changing anything..

brynpickering:
hmm, pre-commit should have fixed this issue when you first committed (assuming you have run `pre-commit install` within your pam environment). Anyway, `pre-commit run -a` should force it to run it on the files you've already committed.

PR #266, opened 2023-12-12
Try updating GDAL

brynpickering:
@Theodore-Chatziioannou was having problems on Windows with GDAL v3.6 but with this relatively strict dependency pinning we might be causing other downstream issues in https://github.com/arup-group/ersa/pull/56 so I'm trying out updating the pinning to skip v3.6 and seeing what the CI thinks about it.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated


brynpickering:
Segementation faults encountered in some Windows runs, so I'll return to this once this issue is resolved: https://github.com/OSGeo/gdal/issues/8959

PR #268, opened 2024-02-23
Link to cookiecutter template

brynpickering:
Brings PAM in line with the latest developments in the [cookiecutter template](https://github.com/arup-group/cookiecutter-pypackage/) that it effectively spawned.

Main change is moving PAM source code into a `src` directory. I know this is something we discussed and dismissed previously @fredshone but work since that time has shown me the benefits of it. I'm now a convert ;)

Also added a CI job that will monitor the template and ensure PAM conforms to any changes made on it in future. 
Should hopefully make it easier to keep PAM up-to-date.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated


brynpickering:
I've updated dependencies to work with Py3.12. Tests seem to pass (maybe you can try it locally on a windows machine too @Theodore-Chatziioannou? GDAL issues _should_ be solved) and I believe the memory profiling test shows a ~20% drop in peak memory consumption (based on another profile that I have from late Sept. 2023).

fredshone:
install and tests run fine on windows 👍🏻

fredshone (approved):

PR #271, opened 2024-04-02
adds note on load time

fredshone:
added comment to workbook as per joss feedback [here](https://github.com/openjournals/joss-reviews/issues/6097#issuecomment-2031440883).


fredshone:
@brynpickering no review required - just a single line of text

PR #278, opened 2024-05-20
Fix the Slack notification channel name in build workflows

### LINKED ISSUE #277: PAM build results are not reported to Slack
## Overview
PAM's daily [CI build workflow](https://github.com/arup-group/pam/blob/main/.github/workflows/daily-scheduled-ci.yml) is scheduled to run at `14:23 UTC` every weekday and should report results to our `#pam-feeds` Slack channel. Although the build [is running as scheduled](https://github.com/arup-group/pam/actions), no messages are being sent to Slack to report the results.

## Futher Detail
The Actions job responsible for sending messages seems to be failing silently - neither the success nor failure job show as complete in the UI:

Screenshot 2024-05-20 at 14 13 20

One reason looks to be an incorrectly configured [Slack channel name](https://github.com/arup-group/pam/blob/0c4f50325ecb24f3db7a7cf8008c3af6463ccd09/.github/workflows/daily-scheduled-ci.yml#L34) (`pam-feed` should be `pam-feeds`). Note that the `docs.yml` workflow[ has the same problem](https://github.com/arup-group/pam/blob/0c4f50325ecb24f3db7a7cf8008c3af6463ccd09/.github/workflows/docs.yml#L34). It is possible that there are multiple problems with this job, but fixing the channel name is a good place to start.

We should fix the broken Slack channel name and possibly also look into failing the build if the notification job fails.

- [ ] macOS
- [ ] Windows
- [X] Linux

### Version

N/A

### Relevant log output

_No response_

mfitz:
- Fixes #277 

The misconfigured channel name my not be the only problem, but it needs fixing in any case and we'll see if the notifications start arriving after that.

### DIFF
--- a/.github/workflows/daily-scheduled-ci.yml
+++ b/.github/workflows/daily-scheduled-ci.yml
@@ -31,5 +31,5 @@ jobs:
     secrets: inherit
     with:
       result: needs.test.result
-      channel: pam-feed
+      channel: pam-feeds
       message: Daily CI action
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -31,5 +31,5 @@ jobs:
     secrets: inherit
     with:
       result: 'failure'
-      channel: pam-feed
+      channel: pam-feeds
       message: Docs Build


mfitz:
We received a Slack message in `#pam-feeds` from the docs build triggered by this branch:

Screenshot 2024-05-20 at 14 37 48

Seems like a god sign!

KasiaKoz (approved):

PR #292, opened 2025-03-26
Fix the daily CI build

### LINKED ISSUE #291: CI Builds are broken
## The Problem

All CI builds are failing due to [an upstream change](https://github.com/arup-group/actions-city-modelling-lab/pull/70/files) in one of our reusable actions.

## The Fix

We should either rename the modified parameters (the values we want to pass remain unchanged), or else pin to an earlier version of the action in question.

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Version

`v0.3.2`

### Relevant log output

```shell
Invalid workflow file: .github/workflows/daily-scheduled-ci.yml#L22
The workflow is not valid. .github/workflows/daily-scheduled-ci.yml (Line: 22, Col: 24): Invalid input, cache_mamba_env is not defined in the referenced workflow. .github/workflows/daily-scheduled-ci.yml (Line: 25, Col: 23): Invalid input, mamba_env_name is not defined in the referenced workflow.
```

mfitz:
- Fixes #291

I've not updated the changelog because this is not a user-facing change.

I used `cruft update` to bring the CI workflows into line with the cookie-cutter template; that was enough to fix at the broken workflows.

## Finding affected files

```shell
grep -iRn "mamba" .

./daily-scheduled-ci.yml:22:      cache_mamba_env: false
./daily-scheduled-ci.yml:25:      mamba_env_name: ${{ needs.get-date.outputs.DATE }}
```

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/.cruft.json
+++ b/.cruft.json
@@ -16,7 +16,7 @@
       "project_visibility": "public",
       "upload_pip_package": "y",
       "upload_conda_package": "y",
-      "upload_aws_image": "n",
+      "upload_aws_image": "y",
       "conda_channel": "city-modelling-lab",
       "command_line_interface": "y",
       "create_docker_file": "y",
--- a/.cruft.json
+++ b/.cruft.json
@@ -1,6 +1,6 @@
 {
   "template": "https://github.com/arup-group/cookiecutter-pypackage.git",
-  "commit": "f0270b09dcc9dd85140b6ad9f146382e1dacdedd",
+  "commit": "2c07547902fe10839dcfb8031ef612bf7ca4f23f",
   "checkout": null,
   "context": {
     "cookiecutter": {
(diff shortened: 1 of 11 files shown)


.cruft.json
```diff
@@ -13,7 +13,8 @@
       "package_name": "cml-pam",
       "module_name": "pam",
       "project_short_description": "The Population activity Modeller (PAM) is a python API for activity sequence modelling.",
-      "upload_pypi_package": "y",
+      "project_visibility": "public",
```
mfitz, line 16: Cruft/cookiecutter needs to know this is a public repo, otherwise `cruft update` complains and errors.

gac55 (approved):

PR #293, opened 2025-03-31
Correct the reusable actions version tags in the CI workflows

### LINKED ISSUE #291: CI Builds are broken

mfitz:
- Fixes #291 

So, the PR I merged to fix the builds used the wrong git tag for some of the actions (`v1.1.1` doesn't exist; we need `v1.1.0` - see [here](https://github.com/arup-group/actions-city-modelling-lab/tags)). Unfortunately, one of the places where the tag was correct was (part of) the CI commit workflow, so the build was fixed for that workflow but broken for the Daily CI, so we only discovered it was broken after merging:

Screenshot 2025-03-31 at 12 37 52

# Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/.github/workflows/commit-ci.yml
+++ b/.github/workflows/commit-ci.yml
@@ -25,5 +25,5 @@ jobs:
   aws-upload:
     needs: test
     if: needs.test.result == 'success'
-    uses: arup-group/actions-city-modelling-lab/.github/workflows/aws-upload.yml@v1.1.1
+    uses: arup-group/actions-city-modelling-lab/.github/workflows/aws-upload.yml@v1.1.0
     secrets: inherit
--- a/.github/workflows/daily-scheduled-ci.yml
+++ b/.github/workflows/daily-scheduled-ci.yml
@@ -13,7 +13,7 @@ jobs:
 
   test:
     needs: get-date
-    uses: arup-group/actions-city-modelling-lab/.github/workflows/python-install-lint-test.yml@v1.1.1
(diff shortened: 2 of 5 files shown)


.github/workflows/commit-ci.yml
```diff
@@ -25,5 +25,5 @@ jobs:
   aws-upload:
     needs: test
     if: needs.test.result == 'success'
-    uses: arup-group/actions-city-modelling-lab/.github/workflows/aws-upload.yml@v1.1.1
+    uses: arup-group/actions-city-modelling-lab/.github/workflows/aws-upload.yml@v1.1.0
```
mfitz, line 28: I used the broken tag here, but the correct tag on line 17, above, so the main linting and testing bit succeeded, but the AWS upload failed

gac55 (approved):
happens to the best of us

